PR #526, opened 2018-07-09
Air 1543

rehanabbasi:
Fix wrongly linkified content on asset page. Updating `Linkify` Regex for links.


rehanabbasi:
@jialin-he I have updated to the linkify regex to accept case-insensitive protocol info. Can you please verify the changes again?

PR #533, opened 2018-07-10
Air-1662

jialin-he:
- Filter assets who have: year = 0, null, 9999
- Display message next to "X results for": "X results have no date and were excluded."
- Fix the sort index showing as an array problem


rehanabbasi:
Looking good!

PR #539, opened 2018-07-13
AIR-1417

jialin-he:
The logic for distinguishing different image group type is:
 - if group_type is 100, it is private
 - if group_type is 200,
    - if owner_id of the group is equal to user's baseProfileId, it is shared
    - otherwise it is institutional
 - if Public = true, it is artstor curated
 - else it is shared with me


src/app/browse-page/card-view/card-view.component.ts
```diff
@@ -89,7 +107,8 @@ export class CardViewComponent implements OnInit {
     this._router.navigate(['/browse','groups', this.browseLevel], { queryParams: queryParams })
   }
 
-  private search_owner(): void {
+  /** Implement the search of owner by owner_id */
+  private searchOwner(): void {
```
codiak: Pass either the group or the owner_id to this function as an argument just to make it clear what this function relies on

Jialin:
Wrong person :)
@jialin-he

PR #541, opened 2018-07-17
AIR-330

rehanabbasi:
Adding `aria-labelledby` attribute to `.meta-blocks`


codiak:
Hey Rehan! Testing with the Mac Voiceover, the metadata is still being skipped. Also, I believe "aria-labelledby" attributes should be on the direct element containing the value, and not the wrapping element which also includes the label

PR #544, opened 2018-07-17
Webpack: Attach cache query param to js files

codiak:
"response-cache-control" query param is attached to the bundled JS files to tell S3 to return a cache control header which will then be picked up by Fastly


rehanabbasi:
Looks good - verified locally the `response-cache-control` param added to the js files `src` by running `yarn build:prod && yarn server:prod`.

PR #545, opened 2018-07-17
Air-1412: Group tags: Display and select tags

jialin-he:
- Selected tags are pushed to the top once selected
 - Add selected tags above the image groups,  and once more than one tag is selected, there is clear all button showing up
 - Only show 15 tags by default and user has a chance to expend it
 - Change the style of the tags according to the design


rehanabbasi:
Looking good!

PR #552, opened 2018-07-23
Air 1404

corbfon:


codiak:
Last bit of feedback: We want the "Groups" title above results to change based on the group type selected (My Groups, Artstor Curated, etc.)

codiak:
Changes were reviewed with Mary. The "All" filter is a perfect solution 🎉 

We also changed the behavior for IP auth to land on the institutional groups. However, the page has a serious problem when loading as a logged in user, which I wasn't able to identify in a timely manner. But there's plenty of logs 😉 

If we can get the browse groups page to load for IP auth, this is good to go (and clear to go to Prod when reviewed)

PR #561, opened 2018-08-02
Air 1705

rehanabbasi:
PACT for groups list endpoint updates.


codiak:
@rehanabbasi Definitely headed in the right direction here, and its well written as a Karma test. However, to have the properties you're verifying be included in the Pact, they needed to be added to the interaction on the "provider". We can meet tomorrow to go over an example I have 👍

PR #570, opened 2018-08-03
AIR-1754

bFraley:


src/app/browse-page/browse-groups/groups.component.pug
```diff
@@ -37,3 +37,4 @@
   br
   div(*ngIf="selectedFilter && selectedFilter.level === 'created'", [innerHtml]="'BROWSE.ERRORS.NO_PRIVATE_GROUPS_COLLECTIONS_PROMPT' | translate")
   div(*ngIf="selectedFilter && selectedFilter.level === 'institution'", [innerHtml]="'BROWSE.ERRORS.NO_INSTITUTIONAL_GROUPS_COLLECTIONS_PROMPT' | translate")
+  ang-access-denied-modal(*ngIf="showAccessDeniedModal")
```
codiak: The modal needs to be moved so it isn't nested inside of the "alert-info" class, it's making the text blue 😮

PR #576, opened 2018-08-07
Air 1751

corbfon:


codiak:
This looks great! One hiccup though... entering Select Mode, you aren't able to click on the image itself to select any more. It's not _bad_, but it is a change in behavior. May want to check with Mary?

PR #577, opened 2018-08-08
AIR-1759

rehanabbasi:
Resolve browse group URLs to properly redirect the old ones.


codiak:
Nice use of a resolver! We could probably do this for other pages too 😄

PR #583, opened 2018-08-10
AIR-1727

jialin-he:
- Add 'private' and 'shared by me' under 'my group'
 - All of the three filters: 'my group', 'private', and 'shared by me' are clickable. When click on 'my group', you'll get the all results (SHARED and UNSHARED) owned by the user
 - The filter is not collapsable


codiak:
Hey Jialin! As "air01@artstor.org", selecting **My Groups** returns 28 pages of results, while **Private** returns 5 pages and **Shared by me** returns 1 page. Shouldn't they equal the total of **My Groups**?

PR #586, opened 2018-08-13
AIR-1756

rehanabbasi:
Add clear search to browse group


rehanabbasi:
Based on the discussion with Kelsey design updates implemented.

rehanabbasi:
@bFraley tooltip styles updated and functionality changed to clear the term even without executing a search. Let me know what you think about the font-family, Previously the tip width and font-size were same as used on the collection-badges, but now I have reduced them to match the designs for clear search tooltip.

PR #588, opened 2018-08-13
AIR-1704

bFraley:
Implements group id pact test


bFraley:
@codiak I considered that, and for `GroupService.get` it makes sense, but the only reason I see for making it separate is to add future tests for an individual group - for the create, copy, delete, and update methods.

Should we go with one pact file for all GroupService methods? I think either way is fine.

codiak:
@corbfon any thoughts on this? I think it would be cleaner in one file, and allow you to reuse the same matchers/expected object. (GET, POST, DELETE... they should all be dealing with objects of the same _form_)

bFraley:
@codiak @corbfon I'm going to make it just one file. Another advantage is when the group api changes, we only need to update the single pact file. Changes on the way...

corbfon:
I agree with the one-file theory here. Response shape sharing will help make it clear to us when there might be different responses expected in different places, and we can remedy that.

codiak:
Thanks for doing extra work on this Brett 👍

PR #594, opened 2018-08-14
Air-1757 (Display number of search results for image group search)

jialin-he:
- Show number of search results and group level when search
 - Remove alert saying you haven't created image group when there is no search result
 - Make sure if there is no result, the pagination will not show "1 of 0"


src/app/browse-page/browse-groups/groups.component.ts
```diff
@@ -337,8 +341,62 @@ export class BrowseGroupsComponent implements OnInit {
...
+          }
+          default: {
+              break
+          }
+        }
+
+        // Set the number of result message
+        if (data.total !== 0){
+          if (this.pagination.size < data.total) {
+            this.numResultMsg = this.pagination.size + ' of ' + data.total + ' results for \"' + this.searchTerm + '\"' + ' from <i>' + groupLabel + '</i>.'
```
rehanabbasi: As per the new design changes from Kelsey - we should be displaying just the number of results instead of number of results out of total number of results for keyword from filter group i.e. `10 results for "Lecture 1" from Institutional groups`.

jialin-he: Just asked Kelsey about this, she conforms that it's better to show `10 of 10 results`. The design is different from the content of the ticket and we agree to follow the ticket.

kelseytrabue: So, I just realized that Jialin's implementation is correct, my design is not. I'm still learning the ropes! Thanks, the design looks good.

rehanabbasi: Okay cool. I think its good to go then. Gonna merge the changes!

PR #601, opened 2018-08-15
AIR-1781: Image Group Search: Return to Page 1 When Search

jialin-he:
Make sure we return to page 1 if we do a search in page larger than 1.


jialin-he:
This brach deals with AIR-1781.

PR #602, opened 2018-08-16
Implement groups.component.cleanGroupSearchTerm

bFraley:
For group search, we need to url encode the term param, as well as replace forward slashes with backslashes


rehanabbasi:
Made a minor update to handle cases where an empty string is passed as a search term. `.replace()` was breaking in cases where the search term was empty. Everything else looks good!

bFraley:
Thanks @rehanabbasi - Good catch on when `term` is empty

PR #608, opened 2018-08-17
AIR-1779 PR#2

bFraley:
GroupSearch.cleanSearchTerm needed all instances of '/' replaced


src/app/browse-page/browse-groups/groups.component.ts
```diff
@@ -489,9 +489,10 @@ export class BrowseGroupsComponent implements OnInit {
    * @returns cleaned and URI encoded string value for the search term
    */
 
-  private cleanGroupSearchTerm(term: string): string {
+  private cleanGroupSearchTerm(term: string) {
```
corbfon: It looks to me like the return type should be left on here

PR #614, opened 2018-08-20
AIR-1787

bFraley:
Update groups.component to just show total, no longer # of total


kelseytrabue:
@bFraley I just left you a comment in Jira I'm not able to see what you're seeing here though :(

PR #630, opened 2018-08-28
Air-1799

rehanabbasi:
Fix `Back to Results` link for image group assets. Using `prevRouteTS` to save the previous route params in session storage.

### DIFF
--- a/src/app/asset-page/asset-page.component.ts
+++ b/src/app/asset-page/asset-page.component.ts
@@ -319,13 +319,17 @@ export class AssetPage implements OnInit, OnDestroy {
                     // Set image share link
                     this.generateImgURL(this.assetIds[0])
                 }
+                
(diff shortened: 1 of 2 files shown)


src/app/asset-page/asset-page.component.ts
```diff
@@ -321,13 +321,15 @@ export class AssetPage implements OnInit, OnDestroy {
                 }
 
                 // For "Go Back to Results" and pagination, for asset that is not from image group look for requestId to set prevRouteParams
-                if (routeParams['requestId'] || routeParams['groupId']) {
+                if (routeParams['requestId'] || routeParams['groupId'] || routeParams['prevRouteTS']) {
```
codiak: Rather than using the timestamp alongside requestId, let's replace requestId's usage. And we should also clean out the "groupId" condition here

PR #638, opened 2018-08-31
Air 1799

rehanabbasi:
Pagination fix for asset page. `prevRouteTS` should be passed with next/prev asset URLs.


bFraley:
@rehanabbasi The new changes resolve all issues, good work!

PR #639, opened 2018-08-31
Asset Page: Patch for Windows download view button UI

bFraley:


bFraley:
-  On MS Browsers, the initial state of the download view button needs to be `enabled` for users to click to init the download

src/app/asset-page/asset-page.component.ts
```diff
@@ -782,23 +785,25 @@ export class AssetPage implements OnInit, OnDestroy {
         .subscribe((blob) => {
           if (blob.size > 0) {
               if (this.isMSAgent) {
-                  this.downloadLoading = false
-                  this.navigator.msSaveBlob(blob, 'download.jpg')
+                  // We give MS Browsers an extra pause of 2 secs
```
corbfon: could you add the reason for why we have the pause? It's apparent from the code that there is a pause, so adding reasoning here would help explain the situation to future code readers

src/app/asset-page/asset-page.component.ts
```diff
@@ -782,23 +785,25 @@ export class AssetPage implements OnInit, OnDestroy {
...
+                  setTimeout(() => {
+                    this.downloadViewReady = true
```
corbfon: Looks like this variable is set multiple times - are there any cases when this might be setting and it's not supposed to?

src/app/asset-page/asset-page.component.ts
bFraley: The extra pause in Windows browsers is there to avoid the file save dialog from triggering before rundownLoadView fully returns

src/app/asset-page/asset-page.component.ts
bFraley: In runDownloadView - it is set to false - then re-enabled just before the file save dialog is triggered.

PR #655, opened 2018-09-13
Attaching "referrer" to /resolve requests

codiak:
We want to provide the referrer information so Binder can begin tracking it


codiak:
A link for testing: http://stage.artstor.org/#/asset/external/8DtZYyMmJloyLyw5ej1%252FRn8r

We want to verify it still works, and its passing a 'ref' property

PR #660, opened 2018-09-17
Air 1829: Safely check index before splicing

codiak:
It was a pretty specific bug that would happen only if "deleted assets" wasn't reset, and the assets didn't exist in the group


src/app/asset-grid/asset-grid.component.ts
```diff
@@ -716,7 +716,11 @@ export class AssetGrid implements OnInit, OnDestroy {
   private removeFromGroup(assetsToRemove: Thumbnail[], clearRestricted?: boolean): void {
     for(let i = 0; i < assetsToRemove.length; i++) {
       let assetId = assetsToRemove[i].objectId
-      this.ig.items.splice(this.ig.items.indexOf(assetId), 1)
+      let igIndex = this.ig.items.indexOf(assetId)
+      // Passing -1 will splice the wrong asset!
```
rehanabbasi: Good catch @codiak :+1:

PR #662, opened 2018-09-18
AIR-1416

rehanabbasi:
Asset Page: CollectionName to IAP and Error Form


bFraley:
@rehanabbasi Nice job!

PR #663, opened 2018-09-18
AIR-1424

bFraley:
Auth Service: add no-cache no-store to refreshUserInstitution

Note: We need to deploy to stage to check if this prevents IE11 from caching institution name from v2/institution


bFraley:
- Verified institution endpoint has unique no-cache param `http://stage.artstor.org/api/v2/institution?no-cache=1537366788822` and `Cache-Control: no-cache, no-store` request header

PR #668, opened 2018-09-20
Add unaffiliated modal for pcollection and image-group-page pages

bFraley:
Add unaffiliated modal for pcollection and image-group-page pages

- We need to check for public users clicking or pasting a link straight to group/id or pcollection/id
pages.


bFraley:
Follows up 1694

PR #670, opened 2018-09-25
AIR-1869

rehanabbasi:
Asset Grid: Fix excluded assets message in date sort for collection, category, pcol, cluster pages.


codiak:
This is looking good! Just one or two places for the code could be condensed in the asset-grid params subscription

rehanabbasi:
@codiak cleaned up some code in asset grid, as we discussed! Mind taking another look at the changes

rehanabbasi:
@codiak cleaned up Asset-Search Service to move the logic for constructing query keyword from `AssetService.queryAll` to `AssetSearchService.search` method

src/app/shared/asset-search.service.ts
```diff
@@ -283,6 +283,19 @@ export class AssetSearchService {
    * @returns       Returns an object with the properties: thumbnails, count, altKey, classificationFacets, geographyFacets, minDate, maxDate, collTypeFacets, dateFacets
    */
   public search(options: SearchOptions, keyword: string, sortIndex): Observable<SearchResponse> {
+    // Construct query keyword based on options
```
codiak: Glad to see this logic getting cleaned up 😄

PR #681, opened 2018-10-01
AIR-1743 (Indicate focus in main nav by more than color)

jialin-he:
- For Chrome and Safari, create an outline around the main nav when focused, which looks like the browser's default focus style.
 - For FireFox, the default outline looks terrible. Specify FireFox's focus style with "-moz-focusring"

### DIFF
--- a/src/sass/modules/_buttons.scss
+++ b/src/sass/modules/_buttons.scss
@@ -156,7 +156,6 @@ $iconBtnSize: $btnHeight;
         
         &:-moz-focusring:not(input):not(select):not(textarea):not(iframe):not(frame):not(body):not(html) {
             outline: 1px dotted;
-            // outline: rgb(77, 144, 254) auto 2px;
         }
     }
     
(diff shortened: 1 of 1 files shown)


src/sass/modules/_buttons.scss
```diff
@@ -132,17 +132,33 @@ $iconBtnSize: $btnHeight;
...
+        text-decoration: none;
+        color: #0039c6;
+        outline: rgb(77, 144, 254) auto 5px;
+        outline-offset: -2px;
+
+        & .value {
+            text-decoration: underline;
+        }
+        
+        &:-moz-focusring:not(input):not(select):not(textarea):not(iframe):not(frame):not(body):not(html) {
```
codiak: I'd like to know more about all these `:not()` selectors 😮

jialin-he: Actually I just get it from line 713 of this document:
https://dxr.mozilla.org/mozilla-central/source/layout/style/res/html.css 
I deleted ":not(button)" for the buttons to be outlined, but I didn't touch others...:sweat_smile:

codiak: Ah, I see! Nice doc reference. We shouldn't need the ":not" selectors since it's being applied to the .btn class.

PR #682, opened 2018-10-02
Air 1349

bFraley:


bFraley:
This PR adds support for the new endpoints:
`allinstitutions`
`alldonatinginstitutions`
`allssinstitutions`

Requests to these are added as a new service named `institutions.service` 
NOTE: The "Contributors List" filter is still behind the contribFilter Flag,
but we use the allinstitutions endpoint to map a user's institution to the contributor count of search results. This enables the Collection Type filter's (type 2) correct count of results to be displayed, and not display the user's institution filter at all (now corrected) if there's not actually results.

PR #699, opened 2018-10-15
AIR-1848 (Make generate citation modal keyboard accessible)

jialin-he:
- Focus on the title of the modal
 - Disable tabbing outside the modal
 - Enable tabbing of the button of Copy-to-Clipboard
 - Change the focus style of help link

Note: Since we use "focusout" event on the last button, when we hit "shift tab" while focusing on the last button, we will go to the title instead of the previous button. This behavior appears on every newly changed modal. Currently haven't figured out a way to solve it.


codiak:
Works well in Chrome and Firefox for me!

PR #707, opened 2018-10-24
AIR-1802

rehanabbasi:
Group Pacts: Update Group by Id PACT to support new items property type


bFraley:
@rehanabbasi Good work on this 👍

PR #713, opened 2018-10-25
Air 1803

rehanabbasi:
Handle new “items” form for /group/GROUPID calls


bFraley:
@rehanabbasi Good job handling this!

PR #717, opened 2018-10-26
Revert "Air 1803"

bFraley:
Reverts ithaka/aiw-ui#713


bFraley:
reverted for commits not pushed

PR #740, opened 2018-11-07
Air-1867 (Error message persists among image group filters)

jialin-he:
- Update both general search component (for groups search) and search component (for search page, collection page and category page).
 - Subscribe to router to remove error message when there are route param changes
 - For general search (group search), disable route param change when user input empty search term to make the above functionality work


bFraley:
Verified SC. @jialin-he +1 for removing template logic to new new method.

PR #760, opened 2018-11-21
AIR-1782 (Firefox image group search - X Icon sizing)

jialin-he:


bFraley:
@jialin-he merging this - but leaving a note here to remind us to check that icon cross all browsers after this change, we may need to use height: auto, but using 1.5 ems fixes the size on FF

PR #775, opened 2018-12-04
Air 1983

bFraley:
NOTE: Branched from 1833

- Uses PLATFORM_ID to define isBrowser boolean, and explicitly checks if localStorage or sessionStorage are defined in a safe way (so they are not referenced on the server in builds)
- Implements Service for localStorage and SessionStorage as an Angular Library
- Provides set / get / remove / clear for localStorage
- Provides set / get / remove / clear for sessionStorage
- Removes use of angular-safeguard, and references to Locker and LockerService
- 'STARTS' to support calls in Idle for progress in SSR


projects/artstor-storage/src/lib/artstor-storage.service.ts
```diff
@@ -0,0 +1,120 @@
...
+import { localStorageFactory } from '@ng-toolkit/universal';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class ArtstorStorageService {
+  constructor(@Inject(PLATFORM_ID) private platformId: Object) {}
+
+  // Is this code being interpretted by a browser-based client?
+  private isBrowser: boolean = this.platformId === 'browser'
```
codiak: Angular does provide isPlatformBrowser(platformId) https://angular.io/api/common/isPlatformBrowser, which I think  we should use consistently, as:
`private isBrowser: boolean = isPlatformBrowser(this.platformId)`

Interestingly, using a simple string comparison should work just fine. PLATFORM_ID should always provide a string output of its token. The only case where it would fail is if in the future we wanted to use service workers, "isPlatformBrowser" would return true for ids of the browser runtime and browser service workers.

codiak:
A decision we should make is how we handle the usage of localStorage as a simple data store. Our code makes assumptions about localStorage always being available and a reliable way of storing a value.

Couple of potential solutions:
1. artstor-storage could fallback to using a local object var (assuming a single instance of the service). Something like: 
`
  if (this.hasLocalStorage()) {
      return localStorage.getItem(key) || this.serviceStorage[key]
   } else {
      return this.serviceStorage[key]
 }
`
2. Similar to the above, but handling it in a wrapping service like we used to with angular safeguard
3. Provide some sort of mocked localStorage to the server app which can act as a simple store

I think 1 and 2 are preferable since we'd have the added benefit of our app recovering from deleted local values (can't think of a way this would happen without the user manually doing so in developer tools though haha)

rehanabbasi:
@bFraley I tried pulling your changes and did 

- yarn
- npm link @angular/cli
- yarn build:ssr:dev && yarn serve:ssr

After disabling JS in chrome I wasn't able to correctly render the asset page (https://localhost:4000/asset/SS33119_33119_15752170) and I see the following logs on node server

```
~ Fresh Request ~
Running on the server with appId=avatar
Constructing app component
Constructing nav component...
Constructing footer...
Running canActivate...
Constructing home component...
Constructing nav component...
ERROR [Error]
~ Fresh Request ~
Running on the server with appId=avatar
Constructing app component
Constructing nav component...
Constructing footer...
Running canActivate...
Constructing home component...
Constructing nav component...
~ Fresh Request ~
Running on the server with appId=avatar
Constructing app component
Constructing nav component...
Constructing footer...
Running canActivate...
Constructing home component...
Constructing nav component...
~ Fresh Request ~
Running on the server with appId=avatar
Constructing app component
Constructing nav component...
Constructing footer...
Running canActivate...
Constructing home component...
Constructing nav component...
~ Fresh Request ~
Running on the server with appId=avatar
Constructing app component
Constructing nav component...
Constructing footer...
Running canActivate...
Constructing home component...
Constructing nav component...
ERROR [Error]
ERROR [Error]
ERROR [Error]
ERROR [Error]
~ Fresh Request ~
Running on the server with appId=avatar
Constructing app component
Constructing nav component...
Constructing footer...
Running canActivate...
Constructing home component...
Constructing nav component...
ERROR [Error]
```

bFraley:
@codiak @rehanabbasi As discussed, I'll be adding an "in-memory" alternative object with methods for get / set and operating on key value storage.

bFraley:
@codiak @rehanabbasi After much trial and error and testing, I think this story is a wrap until we can start executing code of more features, server side.

- I've uncommented Idle logic in AuthService
- Client side all local and session storage is the same as current production
- Server only stores and returns our generic user object

bFraley:
Between this branch and 1987, after breaking canActivate - we're able to move on to a new set of various console and terminal debug errors (a good thing!) for completing this PR and 1987

projects/artstor-storage/src/lib/artstor-storage.service.ts
```diff
@@ -0,0 +1,127 @@
...
+export class ArtstorStorageService {
+  constructor(@Inject(PLATFORM_ID) private platformId: Object) {}
+
+  // Is this code being interpretted by a browser-based client?
+  private isBrowser: boolean = isPlatformBrowser(this.platformId)
+
+  // Server side object to hold generic user object
+  // Note: We don't need one for sessionStorage on the server
+  private localStorageData = this.isBrowser ? null : {
+    user: { data: { status: false, isLoggedIn: false, loggedInSessionLost: false } }
```
codiak: Hmmm, I see what you're doing here, but I don't think we should be handling our generic user object here.

PR #778, opened 2018-12-07
Air 1987

bFraley:
Creating Initial PR for to start review.

PART 1 Work (done)
- Adds DomUtilityService for wrapping DOM based element functions with isBrowser. This also provides us a set of functions to use everywhere as needed.
- Replaces all `document.get... ` references with service calls
- Wraps a few more pieces of logic with isBrowser

TODO:
- Complete inline comments and test service
- **PART 2** - Remove use of calling this service, or document where we can by binding to templates with Element references!


bFraley:
close to open PR for SSR branch, not master!

PR #782, opened 2018-12-12
AIR-1964

bFraley:
Register Comp: Fix Shib messaging
Note: needs deployed to actually test Shib workflow

- Fixes Register form messaging
- Adds Explicit check for 400 status, and that res.code is a Shib error code number


bFraley:
@rehanabbasi I tested these changes on stage and conclude this PR doesn't solve the problem, moving back to in progress.

bFraley:
@codiak closing until work can be continued on this issue.

PR #791, opened 2018-12-19
AIR-2027 (ADL collection titles not displaying on collection page)

jialin-he:
- `/api/category` is no longer available, but `api/v1/categorydesc` returns the category name we need


bFraley:
@jialin We'll leave the old function for now, and remove the commented out code at anther time. Looks good!

PR #792, opened 2018-12-20
AIR-1940

rehanabbasi:
Homepage Accessibility: 
- Adding the missing labels
- Adding role for the nav menu
- Adjusting featured collections animations for SR

### DIFF
--- a/src/app/nav-menu/nav-menu.component.pug
+++ b/src/app/nav-menu/nav-menu.component.pug
@@ -1,14 +1,14 @@
 .cover([class.show]="!mobileCollapsed")
-.container.nav-container
+.container.nav-container(role="navigation", tabindex="1")
   nav.navbar.navbar-expand-sm.navbar-light.bg-faded
     div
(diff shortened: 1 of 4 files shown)


src/app/nav-menu/nav-menu.component.pug
```diff
@@ -1,22 +1,22 @@
...
-.container.nav-container
+.container.nav-container(role="navigation", tabindex="1")
   nav.navbar.navbar-expand-sm.navbar-light.bg-faded
     div
     button.navbar-toggler(tabindex="3", (click)="this.mobileCollapsed = !this.mobileCollapsed", [class.collapsed]="mobileCollapsed", type="button", aria-controls="navbarSupportedContent", aria-expanded="false", aria-label="Toggle navigation")
       span.navbar-toggler-icon
     #navMenu.navbar-collapse.collapse([class.show]="!mobileCollapsed")
-      ul.nav.navbar-nav
-        li.nav-item
+      .nav.navbar-nav
```
bFraley: Are we removing unordered lists / list elements from the nav menu only for screenreader purposes?
I'm suggesting that the semantic meaning of using list elements here is more correct, but we can weigh those options another time. @rehanabbasi @kelseytrabue

bFraley:
Let's leave this branch in place for now...

PR #793, opened 2018-12-20
AIR-1969

rehanabbasi:
Adding keyboard arrow keys for navigation between download button options


bFraley:
Good job patching this one up @rehanabbasi

PR #794, opened 2018-12-20
Air 2028

bFraley:
- Resolves deprecation of `/pccollection` endpoint and now uses Search, tagsService


bFraley:
@codiak @rehanabbasi Leaving this here as a reminder to go back and clean more of the /pccollection logic out (commented out for now)

PR #797, opened 2018-12-26
AIR-2016

rehanabbasi:
Filter ADL collections list by title.


rehanabbasi:
We might also want to clean up (in another ticket) the code for fetching the collections list. Right now every time you type something in the collections filter input field, it executes a search call (identical every time). Ideally, we should call search endpoint only if the collections are not available in the local storage.

PR #814, opened 2019-01-08
AIR-2063

bFraley:
- adlCollectionFilterPipe: add checks for filter and filter length
- check viewId and set searchTerm  to empty for classifcation browse menu item
- Fixes application error when visiting Classification

Search on ADL still fixed, from AIR-2016
ADL ClassificationPage loads without error
Other browse/library routes working
Maintains last search result in params for revisiting Browse ADL search

### DIFF
--- a/src/app/browse-page/library.component.ts
+++ b/src/app/browse-page/library.component.ts
@@ -120,7 +120,7 @@ export class LibraryComponent implements OnInit {
         } else {
           reload = true
         }
-        
+
         // Set searchTerm for pipe when updated
         if (params && params['searchTerm']) {
           this.searchTerm = this.selectedBrowseId === '250' ? '' : params['searchTerm']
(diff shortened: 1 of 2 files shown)


src/app/browse-page/library.component.ts
```diff
@@ -109,113 +109,132 @@ export class LibraryComponent implements OnInit {
...
       this.route.params.pipe(
       map((params: Params) => {
+        let reload: boolean = false
         this.loading = true
 
+        // If view changed, load the page
         if (params && params['viewId']){
-            this.selectedBrowseId = params['viewId'].toString()
-            // this.updateSplashImgURL();
+          reload = (this.selectedBrowseId !== params['viewId'].toString())
```
bFraley: This is the most logical place to handle this... simple :)

PR #816, opened 2019-01-09
AIR-1768

rehanabbasi:
Reload IG after deleting selected assets!


src/app/nav-menu/nav-menu.component.ts
```diff
@@ -214,6 +214,7 @@ export class NavMenu implements OnInit, OnDestroy {
         })
         this._assets.removeFromResults(removeIds, this.ig.items.length) // make the call to asset service which will update the asset grid with modified assets and also pass the total # of items for pagination values
         this._assets.selectModeToggle.emit()
+        this.reloadIG() // Reload IG assets after deleting selected assets
```
bFraley: @rehanabbasi Just taking a quick glance at this PR, not actual review yet. But, I want to just ask for a sanity check here if it's ever possible that `this.reloadIG()` could execute before `this._assets.removeFromResults` completes execution ?

codiak: That's a good point. I think reload should be handled on the subscription

PR #817, opened 2019-01-09
Re-Merge "AIR-2007""

bFraley:
Reverts ithaka/aiw-ui#809


bFraley:
Re-Merge AIR-2007 for deployment

PR #824, opened 2019-01-15
AIR-2041

bFraley:
AuthService, Register: Add navigate to Login from Register

Due to Angular6 upgrade:
We needed to pass true to observer.next in canActivate for public users, so that the register component initializes, where logic for navigating to /login from there is handled.
Otherwise, register is never loaded and the Home component is not rendered.

### DIFF
--- a/src/app/shared/auth.service.ts
+++ b/src/app/shared/auth.service.ts
@@ -460,9 +460,10 @@ export class AuthService implements CanActivate {
         observer.next(true)
       })
     } else if (this.isPublicOnly() && state.url.includes('/register')) {
-      // For unaffiliated users, trying to access /register route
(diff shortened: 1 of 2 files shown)


src/app/shared/auth.service.ts
```diff
@@ -459,11 +459,6 @@ export class AuthService implements CanActivate {
       return new Observable(observer => {
         observer.next(true)
       })
-    } else if (this.isPublicOnly() && state.url.includes('/register')) {
-      // For unaffiliated users, trying to access /register route
-      return new Observable(observer => {
-        observer.next(false)
-      })
```
bFraley: @codiak thanks for this additional change!

PR #832, opened 2019-01-22
Air 1995

bFraley:
- Defines `rights.ts` file with rights object with fields `name`, `link`, 'img'
- Adds rights icons files
- AssetPage changes include:
    - Add `rightsText`, `rightsLink`, `rightsImg` strings
    - Implements `isRightStatement` method


bFraley:
- Added 'hardened' string comparison of rights statements
- Style display text or rights statements

PR #835, opened 2019-01-24
Air 2059: Load asset page server-side for public assets

codiak:


src/app/_services/metadata.service.ts
```diff
@@ -0,0 +1,261 @@
...
+})
+export class MetadataService {
+
+  constructor(
+      private _auth: AuthService,
+      private _http: HttpClient
+  ) { }
+
+
+    public buildAsset(assetId: string, { groupId = '', legacyFlag = true, openlib = false, encrypted = false }={}): Observable<Asset> {
```
bFraley: Do we want to set legacyFlag to true here ?

src/app/app.component.ts
```diff
@@ -22,16 +26,16 @@ import { DomUtilityService } from 'app/shared';
   template: `
     <ang-sky-banner *ngIf="showSkyBanner" [textValue]="skyBannerCopy" (closeBanner)="showSkyBanner = false"></ang-sky-banner>
     <div>
-      <div id="skip" tabindex="-1" aria-activedescendant="button">
+      <div id="skip" tabindex="-1">
         <button id="button" (click)="findMainContent()" (keyup.enter)="findMainContent()" tabindex="1" class="sr-only sr-only-focusable"> Skip to main content </button>
       </div>
       <nav-bar tabindex="-1"></nav-bar>
-    </div>
-    <main tabindex="-1">
```
bFraley: Will this break tabbing order / past accessibility work ? I 'm sure we can review that when looking at merging into stage branch

src/app/app.component.ts
```diff
@@ -53,20 +57,20 @@ export class AppComponent {
...
-      translate.setDefaultLang(langStr);
-      // the lang to use, if the lang isn't available, it will use the current loader to get them
-      translate.use(langStr);
-
+      // Use the translate loader to pull translations client-side
+      this.translate.use(langStr);
+      this.translate.setDefaultLang(langStr);
       // Start GA trackiong
       angulartics2GoogleAnalytics.startTracking()
+    }  else {
```
bFraley: This is great!

bFraley:
@codiak Great work! Let's merge!

src/app/_services/metadata.service.ts
codiak: Good catch...

PR #837, opened 2019-01-25
Air 2002

bFraley:
- Reduce the amount of calls to angular-safeguard from within emitted events / subscriptions
- in `isPublicOnly` don't rely on angular-safeguard at all because we call `isPublicOnly` many times across many templates.
- Cleans up asset-page Tour code no longer needed
- Removes calls to `downloadAuth()` in Asset Page Template and uses new `acceptedTerms` boolean instead.
- Reduces user lookup in localStorage in getUser and saveUser, by mostly checking if we already know the value of this.user and this.user.status in AssetPage

### DIFF
--- a/src/app/shared/auth.service.ts
+++ b/src/app/shared/auth.service.ts
@@ -173,7 +173,7 @@ export class AuthService implements CanActivate {
 
     idle.onTimeout.pipe(
       map(() => {
-        let user = this.user.hasOwnProperty('status') ? this.user : this.getUser();
+        let user = this.getUser()
         // console.log(user);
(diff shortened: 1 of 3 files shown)


src/app/shared/auth.service.ts
```diff
@@ -705,6 +707,8 @@ export class AuthService implements CanActivate {
         'isLoggedIn': false,
         'loggedInSessionLost': loggedInSessionLost
       }
+      // Track whether or not user object has been refreshed since app opened
+      this.userSessionFresh = true
```
codiak: This won't work quite right moved in to this condition...

PR #839, opened 2019-01-28
AIR-2088

bFraley:
Registration:  Increase regex accepted size range of second level domains for long email address domains

- Increased accepted range to 30 chars after name@ in email validation


bFraley:
**Problem:** Our custom email validator regex is too expensive to run on every keypress.
**Solution:** There’s an `updateOn` param that `FormControl` accepts. I set it to ‘blur’, so it will run `emailValidator` once, or whenever input onBlur fires.

PR #843, opened 2019-01-30
Air-2068

bFraley:
Open for initial code review, but needs some debugging of user state.


bFraley:
This PR is now ready for review.
- Remove tour and solrMetadata unused code
- drastically reduce calls to locker service via getUser which is called a lot from isPublicOnly calls in templates
- sets a local this.user var in auth service that gets maintained app wide on user state update routines and events
- mitigates asset page calls to downloadAuth() from template by replacing with `acceptedTerms` var controlled by Output accepted terms in agree modal

src/app/asset-page/asset-page.component.ts
```diff
@@ -333,14 +284,6 @@ export class AssetPage implements OnInit, OnDestroy {
                 if (routeParams && routeParams['featureFlag']) {
                     this._flags[routeParams['featureFlag']] = true
                     this.relatedResFlag = this._flags['related-res-hack'] ? true : false
-                    this.detailViewsFlag = this._flags['detailViews'] ? true : false
```
bFraley: @rehanabbasi I missed this line in resolving conflicts, my bad.

PR #852, opened 2019-02-06
Asset Print: Update to use latest metadata display

codiak:


src/app/asset-pp-page/asset-pp-page.component.ts
```diff
@@ -46,40 +46,41 @@ export class AssetPPPage implements OnInit {
   // Load Image Group Assets
   loadAsset(): void{
     let self = this
-    this._metadata.buildAsset(this.assetId, null).pipe(
+    this._metadata.buildAsset(this.assetId, {}).pipe(
       map(asset => {
-
+        console.log("FIRST LOG", asset, asset.image_compound_urls)
```
codiak:
```suggestion
```

PR #859, opened 2019-02-13
AIR-2093

rehanabbasi:
Asset Page - Design updates for the dropdown buttons.

### DIFF
--- a/src/app/asset-page/asset-page.component.scss
+++ b/src/app/asset-page/asset-page.component.scss
@@ -513,21 +513,28 @@ $titleNavBg: #eee;
 }
 
 // Custom styles for dropdown buttons menu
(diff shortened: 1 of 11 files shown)


src/app/asset-page/asset-page.component.scss
```diff
@@ -511,3 +511,67 @@ $titleNavBg: #eee;
...
+
+        .icon {
+            font-size: 20px;
+            font-weight: 600;
+            margin-right: 5px;
+        }
+    }
+}
+
+.btn.dropdown-toggle {
```
bFraley: @rehanabbasi I added this in the asset scss, because the change made in buttons.scss applied to our nav menu style globally. Leaving this here as a reminder to check that your intended style for padding is good to go on the asset page buttons.

PR #864, opened 2019-02-14
AIR-2133 (Add a loading icon for filter pane on search results page)

jialin-he:
- Replace the nested subscription with two independent subscription. Make sure we assign available filters only when we have both institution result and filters result.
 - Add loading icon when the filters are not loaded.


src/app/asset-filters/asset-filters.component.scss
```diff
@@ -327,4 +327,9 @@ input[type=number]::-webkit-outer-spin-button {
 }
 input[type=number] {
   -moz-appearance: textfield;
+}
+
+.move-up {
```
bFraley: @jialin-he Reminder note to discuss this one next week :)

PR #869, opened 2019-02-19
AIR-2131 (Display "no results" state in groups list when adding item to existing group)

jialin-he:
- Note the change from ">" to ">=" in `add-to-group.component.ts`. I make this change to ensure that when we type in some search term, then delete the search term, the all groups result will reload. (**This is a quick fix not stated in the story.**)
 - Add 'no-result' state when there is no search results
 - Remove the functionality that we only start to search when search term has characters in multiples of 3 (according to Kelsey). So now, once you start typing, your image group results start appearing as you're typing.
 - To prevent overloading the backend when we allow search on every key, add a timeout of 200ms (this means if you type in "abcd" within 200ms, we will only make **one** call to the API with query "abcd" instead of making four calls "a", "ab", "abc", "abcd")
 - Prevent making request with empty search term


src/app/modals/add-to-group/add-to-group.component.ts
```diff
@@ -245,7 +245,7 @@ export class AddToGroupModal implements OnInit, OnDestroy {
   public searchGroups(event?): void {
     // Execute search after every third character of the search term
     // Use ">=" instead of ">" so that when we have empty search term (when we type something and delete it), we will reload the groups
-    if ((this.groupSearchTerm.length >= 0) && (this.groupSearchTerm.length % 3 === 0)) {
```
bFraley: @rehanabbasi @codiak Isn't this so we're not searching on every keypress? Can we think of a different way here?

jialin-he: Kelsey told me yesterday that she wants to search on every keypress. She added this to the ticket.

src/app/modals/add-to-group/add-to-group.component.ts
```diff
@@ -244,13 +246,18 @@ export class AddToGroupModal implements OnInit, OnDestroy {
 
   public searchGroups(event?): void {
     // Execute search after every third character of the search term
```
rehanabbasi: I think we should remove this comment - its kind of misleading with the current implementation!

PR #871, opened 2019-02-21
Air 2119: Generalized Toast service (and Toast stack!) 

codiak:
- Send toasts via toast service
- Display multiple toasts at once
- Allow for disappearing and required-dismissal
- Set time out to 9 sec
- Animate multiple toasts at once


src/sass/core/_mixins.scss
```diff
@@ -0,0 +1,9 @@
+// Mixins for simpler SCSS
+
+@mixin keyframes($name) {
```
rehanabbasi: This is very neat @codiak :ok_hand:

PR #877, opened 2019-02-25
AIR-2090-2

bFraley:
Update Add to Group tabbing functions to handle shift tab from Add button when a selection is made
Also add click event detection to the selection logic to be same as Space orEnter on a group


bFraley:
@rehanabbasi
- Click now also uses the same logic as Space or Enter
- And Shift+Tab from Add button sets focus on the selected IG via it's element id attribute

Let me know if you find any issues, or simply update extra changes here before merge.

PR #879, opened 2019-02-26
Air 2164

bFraley:


bFraley:
- Handles downgraded MV items in metadata service and for loading downgraded MV thumbnail in Asset Print Preview
- An asset that returns an empty `image_compound_urls` array indicates it has been downgraded and the value of `asset.tileSource` and `asset.thumbnailUrl` is set to `storUrl + asset.image_url`'s value

PR #888, opened 2019-02-28
AIR-2141 (Update Image Group Page "Header")

jialin-he:
- A "group-title" component contains new style
 - Change only appearing on image group for now
 - Public css change of search bar with larger margin according to Kelsey
 - Two pseudo-elements for the full-width borders
 - The "Export" and "Present" buttons are just placeholders for now

### DIFF
--- a/src/app/image-group-page/image-group-page.component.pug
+++ b/src/app/image-group-page/image-group-page.component.pug
@@ -4,12 +4,8 @@ div
     .row.no-gutters.row--search-bar
       .col-md-5
         ang-search
-    .row
-      .col-12.pb-4
-        h3#pageType Browse Group
-        .row
-          h1#pageTitle.h1--title-block.col-md-9 {{ ig.name }}
-          
+    ang-group-title([ig]="ig")
+    .row 
       .col-md-12
         ang-asset-grid([actionOptions]="actionOptions", (reordering)="toggleReorder($event)", [igMetaData]="ig")
         .alert.alert-warning(*ngIf!="ig.items && ig.items.length > 0 && !ig.thumbnails")
--- a/src/app/shared/group-title/group-title.component.pug
+++ b/src/app/shared/group-title/group-title.component.pug
@@ -1,11 +1,11 @@
 .row.group-border
-    .col-9.pb-24.pt-24.z-1000
+    .col-9.btn-row
         .row.m-sm-0.pb-3(#titlerow="")
             h1.title {{ ig.name }}
         .row.m-sm-0.item-last-updated
             span {{ ig.total }} items
             span.dot •
             span Last updated {{ ig.update_date | date:'MMMM dd, yyyy' }}
-    .col-3.pb-24.pt-24.z-1000
+    .col-3.btn-row
         button.btn.btn-primary.ml-3 Export
         button.btn.btn-primary.ml-3 Present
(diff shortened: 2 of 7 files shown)


src/app/shared/group-title/group-title.component.pug
```diff
@@ -0,0 +1,11 @@
+.row.group-border
+    .col-9.pb-24.pt-24.z-1000
```
bFraley: @jialin-he I'm curious about the z index 1000 here.

src/app/shared/group-title/group-title.component.pug
```diff
@@ -0,0 +1,11 @@
...
+        .row.m-sm-0.pb-3(#titlerow="")
+            h1.title {{ ig.name }}
+        .row.m-sm-0.item-last-updated
+            span {{ ig.total }} items
+            span &nbsp;&nbsp;•&nbsp;&nbsp;
+            span Last updated {{ ig.update_date | date:'MMMM dd, yyyy' }}
```
bFraley: +1 For using date format pipe

src/app/shared/group-title/group-title.component.pug
```diff
@@ -0,0 +1,11 @@
+.row.group-border
+    .col-9.pb-24.pt-24.z-1000
+        .row.m-sm-0.pb-3(#titlerow="")
+            h1.title {{ ig.name }}
+        .row.m-sm-0.item-last-updated
+            span {{ ig.total }} items
+            span &nbsp;&nbsp;•&nbsp;&nbsp;
```
bFraley: I'd like @codiak 's approval here on using no break space, instead of CSS. This works for me though 👍

src/app/shared/group-title/group-title.component.pug
jialin-he: If I don't set a higher z-index here, the content will be covered by the pseudo element created by ":after" in the css. (And therefore you cannot select the words with your mouse.)

src/app/shared/group-title/group-title.component.pug
codiak: Oooh, a class for the dot would be much better. Something inline-block with a margin 🙏

src/app/shared/group-title/group-title.component.pug
```diff
@@ -0,0 +1,11 @@
...
+            span Last updated {{ ig.update_date | date:'MMMM dd, yyyy' }}
+    .col-3.pb-24.pt-24.z-1000
```
codiak: Rather than using padding/margin helpers, we have a handy ".btn-row" class. One day we'll have styleguide that include all these utilities 😂 
But for now, if you could use:
.btn-row
     .btn
     .btn 
Your spacing should be handled for you

src/app/image-group-page/image-group-page.component.pug
```diff
@@ -4,12 +4,8 @@ div
     .row.no-gutters.row--search-bar
       .col-md-5
         ang-search
-    .row
-      .col-12.pb-4
-        h3#pageType Browse Group
-        .row
-          h1#pageTitle.h1--title-block.col-md-9 {{ ig.name }}
-          
+    ang-group-title([ig]="ig")
```
codiak: We should probably ngIf this to only load once the ig is loaded. Right now the half loaded header is a little rough

jialin-he:
Have made some changes according to @codiak 's suggestion 👌

PR #891, opened 2019-03-01
AIR-2166 (AIW and Admin: New error code for passwords we can't match)

jialin-he:
- Check for error code 422 for lost password
 - Update error message for lost password

Note: the message in the response seems to have NOT been decided by Auth team. It may be easier for us to check the response message later?


bFraley:
Pending service support for new error code...

PR #892, opened 2019-03-01
AIR-2056 (Collection type filter is empty on search page for some institutional collections)

jialin-he:
- Assign correct count to collection type filters so that the collection type filter is not empty
 - Remove filter if there is only one facet in it and the only facet's name is empty

### DIFF
--- a/src/app/asset-filters/asset-filters.component.ts
+++ b/src/app/asset-filters/asset-filters.component.ts
@@ -198,7 +198,7 @@ export class AssetFilters {
 
         // The loggedd in user's institutional 'Collection Type' filter
         if (colType.name == '2' || colType.name == '4') {
-          filters['collectiontypes'][i]['count'] = this.instFilterCount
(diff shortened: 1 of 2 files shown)


src/app/asset-filters/asset-filters.component.ts
```diff
@@ -198,7 +198,7 @@ export class AssetFilters {
 
         // The loggedd in user's institutional 'Collection Type' filter
         if (colType.name == '2' || colType.name == '4') {
-          filters['collectiontypes'][i]['count'] = this.instFilterCount
```
codiak: `instFilterCount` is explicitly set for this purpose because it should reference the `donatinginstitutionids` count

jialin-he: Sorry I didn't fully understand the requirement, now it's back. The purpose now is to remove the filter header if it only contains one facet and the name of the facet is empty or the count is 0. New changes have been made~

PR #898, opened 2019-03-07
AIR-1810: (Save Detail to Group)

bFraley:
Add To Group and Group Service - for saving zoomed detail

- Adds `groupV2 ` `group.service`
- Map group.items array of strings `["item 1 id", "item 2 id"]` to map array of item objects, as `["id": "item id"...]`
- An asset with zoom details includes the `zoom {}` object for updating a group


bFraley:
Additional changes:
- Group Service: define group APIs as groupV1 and groupV2  …
- Add To Group Comp: Define ZoomDetails interface with all number type properties
- Ensure group update for zoomed detail saves integer zoom values

bFraley:
@codiak @rehanabbasi This PR is ready.

Note: Follow up story for refactoring Group interfaces and services is created, instead of much more larger refactoring for this PR.

Outstanding Issues for feature:
- Select Mode changes are needed
- Cardview and IG Modal list thumbnails for detail views
- Print Preview and IG Print Preview thumbnails for detail views

src/app/shared/group.service.ts
```diff
@@ -289,4 +303,25 @@ export class GroupService {
...
+      return zoom
+    }
+}
+
+interface ZoomDetails {
+  viewerX: number
+  viewerY: number
+  pointWidth: number
+  pointHeight: number
+  index: number
```
codiak:
```suggestion
  index?: number
```

PR #902, opened 2019-03-11
AIR-2202

rehanabbasi:
Show Ethnio survey.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -124,6 +124,15 @@ export class AppComponent {
             this.resetOgpTags();
           }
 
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -124,6 +124,15 @@ export class AppComponent {
             this.resetOgpTags();
           }
 
+          // Show Ethnio survey on asset, search, collection, browse-groups, group routes
```
bFraley: @rehanabbasi I'm not sure we've used this approach before, as opposed to calling load script per component. I like it!

PR #916, opened 2019-03-18
FIX: When deleting from group, check against new item format

codiak:


src/app/nav-menu/nav-menu.component.ts
```diff
@@ -211,7 +211,8 @@ export class NavMenu implements OnInit, OnDestroy {
     putGroup.items = putGroup.items.filter((item) => {
       assetFound = false
       this._assets.getSelectedAssets().forEach((asset) => {
-        if (asset.objectId == item) {
+        // Support both legacy and new Group items format
+        if (asset.objectId === item || asset.objectId === item.id) {
```
bFraley: @codiak nice and simple 👍

PR #920, opened 2019-03-19
Air 2170: Use Google Tag Manager

codiak:
- Track existing events via Google Tag Manager


src/app/login-form/login-form.component.ts
```diff
@@ -70,13 +70,6 @@ export class LoginFormComponent implements OnInit {
 
   loadForUser(data: any) {
     if (data && data.user) {
-      data.user.hasOwnProperty('username') && this.angulartics.setUsername.next(data.user.username);
```
bFraley: @codiak This is a good decision to move to auth.service.

PR #921, opened 2019-03-19
Air 2081 (style updates)

bFraley:


bFraley:
@rehanabbasi Includes just style updates. 2nd PR for Slides hookup to button will follow.

PR #924, opened 2019-03-21
AIR-2217 (Design Refine: toast fix, text box being pushed below icon)

jialin-he:
- Set a max-width to prevent text div being push to next line
 - Make toast smaller when the window is small (According to Kelsey, the width is 280px)
 - Wrap long word


codiak:
Looks really good ✨ 
Only one small tweak was needed for multiple toasts on mobile (they were showing on top of each other with position absolute).

PR #925, opened 2019-03-21
AIR-2218 (Disable add detail if viewer doesn't load)

jialin-he:
- Disable add detail to group when viewer doesn't load


codiak:
Clean solution @jialin-he  ✨

PR #933, opened 2019-03-26
AIR-2191 Modals: Show session expired modal first, add priority classes

codiak:
Praveena and I believe the users situation is caused by not giving priority to the Session Expired modal, which allowed the user to try to continue to download or add items while their session was expired

### DIFF
--- a/src/sass/modules/_modals.scss
+++ b/src/sass/modules/_modals.scss
@@ -11,6 +11,18 @@
     background-color: rgba(0, 0, 0, 0.55);
     z-index: 9000;
     overflow-y: auto;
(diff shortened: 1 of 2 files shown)


src/sass/modules/_modals.scss
```diff
@@ -11,6 +11,18 @@
     background-color: rgba(0, 0, 0, 0.55);
     z-index: 9000;
     overflow-y: auto;
+
+    // Currently reserved for the session expired modal
```
bFraley: @codiak This is an interesting and elegant solution!

PR #934, opened 2019-03-28
AIR-2239

rehanabbasi:
* Make sure deleting a detail / original item doesn't remove the items with same ids.
* Make sure we are displaying the items within the start & end index as per the pagination values.

### DIFF
--- a/src/app/shared/assets.service.ts
+++ b/src/app/shared/assets.service.ts
@@ -818,7 +818,7 @@ export class AssetService {
                         let results = res
                         data.thumbnails = results['items'] // V1 ?items from search
                         // For multi-view items, make the thumbnail urls and update the array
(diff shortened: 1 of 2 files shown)


src/app/shared/assets.service.ts
```diff
@@ -818,7 +818,7 @@ export class AssetService {
                         let results = res
                         data.thumbnails = results['items'] // V1 ?items from search
                         // For multi-view items, make the thumbnail urls and update the array
-                        data.thumbnails = data.items.map((item) => {
+                        data.thumbnails = data.items.slice(pageStart, pageEnd).map((item) => {
```
codiak: Worth noting that "slice" is a nice function since its failure state if items = [] would be to just return []

PR #936, opened 2019-03-28
Air 2191: Unique session expired cases

codiak:
+ Catch "null" values from download service
+ 2x Error handling for PUT/POST situation

### DIFF
--- a/src/app/shared/group.service.ts
+++ b/src/app/shared/group.service.ts
@@ -240,17 +240,27 @@ export class GroupService {
             putGroup,
             this.options
         )
-        .pipe(catchError(err =>
-            this.http.post(
-                // Backup POST call for PUT failures over Proxies
-                reqUrl,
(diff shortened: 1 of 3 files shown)


src/app/shared/group.service.ts
```diff
@@ -240,17 +240,27 @@ export class GroupService {
...
-                {
-                  headers: new HttpHeaders({ 'Accept' : 'application/json', 'X-HTTP-Method-Override' : 'PUT' }),
-                  withCredentials: true
+        .pipe(
+            catchError(
+                err => {
+                    if (err.status === 403) {
+                        // Unauthorized/session expired - we assume expired since we don't allow modifying read-only groups
+                        // throw err
+                        return of(err)
```
bFraley: @codiak we need `import { of } from 'rxjs/operators' here

PR #938, opened 2019-03-29
AIR-2198

rehanabbasi:
Browse groups card view - thumbnail image states!

### DIFF
--- a/src/app/browse-page/card-view/card-view.component.ts
+++ b/src/app/browse-page/card-view/card-view.component.ts
@@ -66,15 +66,15 @@ export class CardViewComponent implements OnInit {
       this._assets.getAllThumbnails(itemIds)
         .then( allThumbnails => {
           this.thumbnails = allThumbnails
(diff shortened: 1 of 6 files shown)


src/app/browse-page/card-view/card-view.component.ts
```diff
@@ -63,20 +65,25 @@ export class CardViewComponent implements OnInit {
...
-          })
-
-          this.thumbnails.forEach(thumbnail => {
-            if (thumbnail['media']) {
-              thumbnail.thumbnailImgUrl = thumbnail.media.thumbnailSizeOnePath
-            }
-          })
+          this.thumbnails = allThumbnails
+          // this.thumbnails = allThumbnails.filter((thumbnail) => {
+          //   return thumbnail.status === 'available'
```
codiak: I'm pretty sure we should still be checking for status? As well as thumb path below?

codiak: @bFraley

PR #941, opened 2019-04-02
Air 2240: Ghost reset modal

codiak:
You can test with :
qatester@artstor.org / artstor
BUT DO NOT RESET THE PASSWORD!!!!!!!


bFraley:
@codiak Approved.. will merge after accessibility updates. Am I missing any other additional changes being added?

codiak:
@bFraley the initial focus and tabbing to the support link still need to be fixed

PR #958, opened 2019-04-11
AIR-2248 (Analytics/GTM: Events with missing information [PLACEHOLDER])

jialin-he:
- The labels are not showing in Analytics because of a wrong setting on the TagManager website. So setting a correct variable on TagManager does the fix once for all.
 - This branch just fixes the label for copyGroup where it is undefined.


bFraley:
Note: This could be related to having to reload tab after an image group update - where prepareGroup doesn't define a group.id and group.id is only defined in `if editIG`

PR #959, opened 2019-04-11
AIR-2178 (Make toasts accessible)

jialin-he:
- Set up a generalized live region in index.html
 - Assign content to aria live region when toast loads (Note: I don't put them in a function because it will cause a delay for the VoiceOver to read out the content of the live region.)
 - Remove content from aria live region when the toast is gone (Note: I don't put this in ngOnDestroy because it will be too late and the screen reader will read it on the other page.)
 - Set key board shortcuts: ctrl + g to go to the link, and ctrl + x to close the toast
 - Pause the timer if there is focus in the toast
 - Extend the length of toast to 7 seconds to makes sure the screen reader has enough time to read the notification
 - Add aria label to the close button in the toast
 - Set correct tab flow: 
     - After hit `Add` button in the add to group modal, next thing to focus is the link in the toast
     - After close the toast, focus on `Add to group` dropdown


src/app/shared/toast/toast.component.ts
```diff
@@ -79,6 +110,27 @@ export class ToastComponent implements OnInit, OnDestroy {
     }
   }
 
+  pauseTimer() {
+    this._toasts.cancelToastTimer(this.toast.id)
+  }
+
+  resumeTimer(event: any) {
```
bFraley: When visiting the Go to group page link, or another route, `this._dom.byId("addToGroupDropdown").focus()` will result in a console error.

The very short timeout amount, along with checking for activeAlement is somewhat brittle. Let's talk and brainstorm some options for different solutions.

src/app/shared/toast/toast.component.ts
```diff
@@ -36,17 +40,44 @@ export class ToastComponent implements OnInit, OnDestroy {
     this.stringHTML = this.toast.stringHTML
     this.type = this.toast.type
     this.links = this.toast.links
+    this.plainText = this.stringHTML.replace(/<(?:.|\n)*?>/gm, '')
```
bFraley: Let's look at alternatives / pre-built functions available for stripping HTML tags here, when defining the plain text string.

bFraley: although, this is sort of nitpick.. we shouldn't be performing low level functionality (like stripping html tags) at such a high level in our application

PR #970, opened 2019-04-17
AIR-2281

rehanabbasi:
Add to group modal - show error toast when adding to a group that already contains 1000 or more items!


codiak:
Nicely done 👍 Pretty sure toasts having the highest z-index will always make sense, but we should keep an eye out for cases we might want them to behave otherwise

PR #972, opened 2019-04-17
AIR-2283(Thumbnails for detailed views of downgraded multi view images are not loading in the group)

bFraley:
Thumbnail Comp: Add conditions for Detail Views that are also Downgraded Media


bFraley:
detailViews use `makeDetailViewThmb` for thumbnail src urls, but conditional for the case of `isDetailView && isDowngradedMedia` needed to be added, and uses `stor + thumbnailImgUrl ` -  the same way that multiview and downgraded from MV thumbnails

PR #977, opened 2019-04-19
Air-2124

bFraley:
- Adds `collections/id`, `/categorynames/id` Pact tests
- Moves `getCollectionInfo` and `getCategoryNames` to `collection.service`
- defines Pact mock responses and expected keys data below all test scenarios so they're grouped in one place in the file.

**NOTE:** PR is in progress, pending institutions endpoint pacts...


src/app/_services/collection.service.ts
```diff
@@ -0,0 +1,94 @@
...
+   */
+  /**
+  * Get metadata about a collection
+  * @param colId The collection ID
+  */
+  public getCollectionInfo(colId: string) {
+    let options = { withCredentials: true };
+
+    return this.http
+      .get(this._auth.getUrl() + '/api/v1/collections/' + colId, options) // @TODO why is adding '/api/' needed for Pact call?
```
bFraley: needs fix before merge

src/app/_services/collection.service.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Get description metadata about a Category
+   * @param catId The Category ID
+   */
+  public getCategoryDescription(catId: string) {
+    let options = { withCredentials: true };
+
+    return this.http
+      .get(this._auth.getUrl() + '/api/v1/categorydesc/' + catId, options) // @TODO why is adding '/api/' needed for Pact call?
```
bFraley: needs fix before merge

bFraley:
@ithaka/ang-air-ui This one is all set!

codiak:
Looks like I'm getting only one Collection Pact error:
```
Collections #pact #collections
    GET /api/v1/categorydesc
      ✖ should return a category description response
        Chrome 73.0.3683 (Mac OS X 10.13.6)
      Error: Encountered undefined provider! Usually this means you have a circular dependencies. This might be caused by using 'barrel' index.ts files.
      error properties: Object({ ngSyntaxError: true })
          at syntaxError (http://localhost:9876/node_modules/@angular/compiler/fesm5/compiler.js?:2430:1)
```

codiak:
Once that's resolved, I can manually publish so we don't push failing Groups pacts to the broker 👍

PR #978, opened 2019-04-19
AIR-2295: Add cache headers to /assets and js bundles

codiak:


bFraley:
@codiak This is solid! Let's try it out!

PR #988, opened 2019-04-25
Air 2148 b: Edit group modal tweaks

codiak:
- Prevent background tabbing for Edit group modal 
- Change the button text in the modal from "Close" to "Cancel"


src/app/modals/new-ig-modal/new-ig-modal.component.ts
```diff
@@ -147,10 +141,32 @@ export class NewIgModal implements OnInit {
     }
   }
 
+  ngAfterViewInit() {
```
bFraley: @codiak This is a great approach and pattern to handling initial modal focus across all modals!

PR #995, opened 2019-05-01
Edit Personal Collection image preview fix

bFraley:
- EditPC-Comp: change image preview height to auto instead of 100

Screen Shot 2019-05-01 at 12 29 37 PM


bFraley:
We need this to fix a change in Chrome 74

codiak:
This might be one of our "cooler" bugs...

PR #996, opened 2019-05-02
Air 2121: Download details and full images from multiviews

codiak:


src/app/asset-page/asset-page.component.ts
```diff
@@ -1218,6 +1228,16 @@ export class AssetPage implements OnInit, OnDestroy {
         }
     }
 
+    /**
```
bFraley: @codiak good idea here 👍

src/app/asset-page/asset-page.component.ts
```diff
@@ -617,6 +617,11 @@ export class AssetPage implements OnInit, OnDestroy {
      */
     setDownloadView(): void {
         this.downloadUrl = this.downloadViewLink
+        if (this.assetGroupId && this.downloadUrl.indexOf("/media/") === -1 ) {
```
bFraley: Is this the only case we need to ignore?

PR #997, opened 2019-05-02
AIR-1973

bFraley:
- AssetPage: Remove navigation to empty asset route because this now breaks with SSR requests to `/asset`.
- Increase timeout of page reload to 1sec after Edit details form is saved


bFraley:
- Includes corresponding Ruby test suite pull request https://github.com/ithaka/ang_ui_tests/pull/432

PR #999, opened 2019-05-03
AIR-1227

rehanabbasi:
Advance Search Modal: 
* Apply `OR` operator between filters within the same filter group
* Filters from multiple filter groups should be concatenated with an `AND` operator.


rehanabbasi:
@codiak I have made the relevant changes to load the applied filters correctly once the user opens the advance search modal for edit!

PR #1017, opened 2019-05-15
AIR-2338(Add mouseflow)

bFraley:
ScriptStore: Adds mouseflow script
Search page: bind _mfq to window for mouseflow

Mouseflow call from localhost on search page:

Screen Shot 2019-05-15 at 11 18 51 AM

### DIFF
--- a/src/app/shared/script.service.ts
+++ b/src/app/shared/script.service.ts
@@ -59,6 +59,11 @@ export class ScriptService {
                         resolve({script: name, loaded: true, status: 'loaded'});
                     };
                 }
+                // window binding for mouseflow script
+                if (script.id === 'mouseflow') {
+                  script.innerText = 'window._mfq = window._mfq || [];'
+                }
+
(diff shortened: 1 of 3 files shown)


src/app/shared/script.service.ts
```diff
@@ -59,6 +59,11 @@ export class ScriptService {
                         resolve({script: name, loaded: true, status: 'loaded'});
                     };
                 }
+                // window binding for mouseflow script
+                if (script.id === 'mouseflow') {
+                  script.innerText = 'window._mfq = window._mfq || [];'
```
codiak: Nice solution! What if we moved the innerText value to the "store" object, so any scripts in the future could include an innerText?

bFraley: Really good idea! Update coming...

bFraley: @codiak Updated. Let me know what you think 👍

PR #1025, opened 2019-05-21
Air 2338: Mouseflow follow up 

codiak:
It looks like page views have to be manually reported in an SPA


src/app/search-page/search-page.component.ts
```diff
@@ -60,7 +60,15 @@ export class SearchPage implements OnInit, OnDestroy {
 
     // Mouseflow script
     this._script.loadScript('mouseflow')
-
+      .then(() => {
+        if (window['_mfq']) {
+          // Trigger mouseflow page view
+          // Doc: http://help.mouseflow.com/knowledge_base/topics/single-page-websites
+          window['_mfq'].push(["newPageView", "/search"])
```
bFraley: 👍

PR #1034, opened 2019-05-29
Air 2306

bFraley:


bFraley:
Adds `user-register.pact.ts` which tests:
 - successful registration
- already registered
- invalid input (Bad Request)

Changes `auth.service` to use `environment.API_URL` instead of `getHostname` and `getUrl`

**NOTES**
 - Docs for error message responses are inconsistent, but the value of `invalidInputResponse` is still used in test until we determine that messaging is not something we receive or care to directly test for. Instead we catch and test the 400 Bad Request response (not testing the value of the actual interaction's response body because we don't receive that)

- To be determined, is the form values to be used for creating a user successfully, along with the teardown and removal of that user.

- To be determined, is the contract definition and alignment of how we test other forms of registration (Shibboleth)

PR #1035, opened 2019-05-30
AIR-2362

rehanabbasi:
Changes to new registered account email checkboxes.
Collapse the first two checkboxes into a single checkbox to receive updates from ARTstor.


rehanabbasi:
@codiak changes made to make sure the data is sent with old values by setting both `info` & `survey`.

PR #1039, opened 2019-06-04
Air 2333: Abstracted thumbnail type: AssetThumbnail

codiak:
When reviewing, pay attention to how AssetThumbnail is passed around, and make sure its typed where it should be. Also, check variants of asset:
- Zoomed details
- Multi views
- Docs
- Video/audio
- Groups vs search


src/app/asset-grid/thumbnail/thumbnail.component.pug
```diff
@@ -1,27 +1,16 @@
-.card-block.card-text(tabindex="3", [attr.aria-label]="thumbnail.tombstone && thumbnail.tombstone.length > 0 ? thumbnail.tombstone.join(', ') : thumbnail.name + ', '  + thumbnail.agent + ', ' + thumbnail.date", [attr.id]="('item-' + itemIndex)", [ngClass]="{ 'reorder--asset' : arrowReorderMode }")
```
bFraley: @codiak Can you please review this line? I removed the check for `thumbnail.tombstone` and this whole ternary for the aria label. The `thumbnail.tombstone ` will throw property does not exist error on prod build.

src/app/_services/thumbnail.service.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  private getCollectionType(thumbnail: any): CollectionTypeInfo {
+    let collectionValue = thumbnail['collectionType'] ? [ thumbnail['collectionType'] ] : thumbnail['collectiontypes']
+    return CollectionTypeHandler.getCollectionType(collectionValue, thumbnail['contributinginstitutionid'])
+  }
+
+  /**
+   * Get image url to set on thumbnail
+   */
+  getThumbnailImg(thumbnail: AssetThumbnail): string {
```
bFraley: Do we want to stick this with the other public methods?

codiak: This one should actually be private

codiak: Just kidding! You're right

PR #1047, opened 2019-06-13
Air 2334: Resolve Googlebot loading errors

codiak:
+ Disable js for bots on server-rendered pages
+ Return 503s if asset is not rendered
+ Attach canonical header if asset is rendered

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "domino": "^2.1.1",
     "driver.js": "^0.8.1",
     "ie-shim": "^0.1.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -103,6 +103,7 @@
     "domino": "^2.1.1",
     "driver.js": "^0.8.1",
     "ie-shim": "^0.1.0",
+    "isbot": "^2.3.0",
```
bFraley: I'm liking where this going already :)

PR #1052, opened 2019-06-21
AIR-2354

jialin-he:
- Cause: the text in the button is overriding the visiblity of the links id's (exportPptLink, exportZipLink)
 - Solution: Give both parent and child element an id start with "exportPpt" (for example, the parent element is called "exportPptLink", and the child elements are called "exportPptIcon" and "exportPptContent"). Then in the Tag Manager setting, we can set the Click to be triggered by Click ID starts with "exportPpt".
 - For test: Tag Manager preview link of the new change: https://www.googletagmanager.com/start_preview/gtm?uiv2&id=GTM-TMVNHFH&gtm_auth=C06s3pTeBCJh_owwOJafRA&gtm_preview=env-25&gtm_debug=x


bFraley:
@jialin-he  Clean, and super clever way to handle this!

PR #1059, opened 2019-06-25
AIR-2415

jialin-he:
- Add feature flag to the flag service (newExport)
 - When there is newExport feature flag, append "?springBoot=true" to the url

### DIFF
--- a/src/app/_services/image-group.service.ts
+++ b/src/app/_services/image-group.service.ts
@@ -83,7 +83,7 @@ export class ImageGroupService {
   }
 
   /** Gets the link at which the resource can be downloaded. Will be set to the "accept" button's download property */
-  public getDownloadLink(group: ImageGroup, zip ?: boolean): Promise<any> {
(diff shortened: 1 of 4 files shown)


src/app/_services/image-group.service.ts
```diff
@@ -119,10 +119,14 @@ export class ImageGroupService {
       return data
     })
     .then((data) => {
+      let groupExportUrl = url + '/' + format + '/' + group.id + '/' + useLegacyMetadata
+      groupExportUrl = newExport ? groupExportUrl + '?springBoot=true' : groupExportUrl
+
       // Return request that provides file URL
       return this.http
-        .post(url + '/' + format + '/' + group.id + '/' + useLegacyMetadata, this._auth.formEncode(data), options)
+        .post(groupExportUrl, this._auth.formEncode(data), options)
```
bFraley: Thank you for setting this as a variable!

PR #1063, opened 2019-06-27
AIR-2423

bFraley:
- newExport
- newExportRollout for CA 🇨🇦

### DIFF
--- a/src/app/_services/flag.service.ts
+++ b/src/app/_services/flag.service.ts
@@ -35,26 +35,20 @@ export class FlagService {
     ).pipe(
     map(res => {
       let flags = res.body
-      let userCountryCode: string = res.headers.get('x-artstor-country-code').substr(0, 2)
 
-      /**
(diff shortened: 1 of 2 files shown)


src/app/_services/flag.service.ts
```diff
@@ -35,26 +35,20 @@ export class FlagService {
...
-       * EXAMPLE USE OF COUNTRY CODE ROLLOUT
-       * // Boolean assignments
-       * this.unaffiliated = flags.unaffiliatedAccess
-       *
-       * // If the user's country code is allowed, set unaffiliated flag to true
-       * let userCountryCode: string = res.headers.get('x-artstor-country-code').substr(0, 2)
-       * if (flags.unaffiliatedAccessRollout.indexOf(userCountryCode) > -1) {
-       *  this.unaffiliated = true
-       * }
-       */
```
codiak: We should leave the commented example in, for future use? Makes clean up later easier

PR #1064, opened 2019-06-27
AIR-2366

jialin-he:
- Captain's log: log request iap event
 - This new event is documented in this wiki page: https://wiki.jstor.org/display/ANG/Captain%27s+Log+for+AIW+Frontend


codiak:
Nice solution, and good job avoiding logging within a function that binds to a "href" value... I've made that mistake before

PR #1080, opened 2019-07-08
Air 2307

bFraley:
Adds logout interaction to UserAccess Pact

### DIFF
--- a/src/app/pact/user-register.pact.spec.ts
+++ b/src/app/pact/user-register.pact.spec.ts
@@ -58,7 +58,7 @@ describe('Register form POST /api/secure/register #pact #user-register', () => {
           uponReceiving: 'registration form submission from a new user',
           withRequest: {
             method: 'POST',
-            path: '/api/secure/register',
+            path: '/api/secure/register?' + _auth.getAuthLogParams(),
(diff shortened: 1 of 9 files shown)


src/app/pact/user-register.pact.spec.ts
```diff
@@ -58,7 +58,7 @@ describe('Register form POST /api/secure/register #pact #user-register', () => {
           uponReceiving: 'registration form submission from a new user',
           withRequest: {
             method: 'POST',
-            path: '/api/secure/register',
+            path: '/api/secure/register?' + _auth.getAuthLogParams(),
```
bFraley: This doesn't entirely fix updates needed in user-register pact, but a start. I can remove updates to this file, in this PR if needed, but it doesn't hurt? @codiak

bFraley:
All 3 Passing in User Access 👍

src/app/pact/user-register.pact.spec.ts
codiak: It might be better to not test/include tracking params within Pacts. I wonder if we could just have "getAuthLogParams" return an empty string in the test environment?

PR #1083, opened 2019-07-10
AIR-2396 (Captain's log event: Zoom on image)

jialin-he:
- We auto zoom when the viewer is first loaded, track the zoom event after the auto zoom is finished
 - Make sure the event is logged once per user per asset per session

### DIFF
--- a/src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
+++ b/src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
@@ -11,6 +11,7 @@ import { AuthService, LogService, ToolboxService } from '_services'
 import { Asset, ImageZoomParams } from 'datatypes'
 import { MetadataService } from 'app/_services'
 import { isPlatformBrowser } from '@angular/common'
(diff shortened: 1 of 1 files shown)


src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
```diff
@@ -423,6 +426,18 @@ export class ArtstorViewerComponent implements OnInit, OnDestroy {
...
         this.osdViewer.addHandler('zoom', (value: any) => {
+            if (this.autoZoomFinished && value.eventSource.id !== this._storage.getSession('loggedZoomAsset')) {
+                this._log.log({
+                    eventType: 'artstor_zoom',
+                    additional_fields: {
+                        item_id: this._assetId,
+                        fullscreen: this.isFullscreen
+                    }
+                })
+                this._storage.setSession('loggedZoomAsset', value.eventSource.id)
```
codiak: While it might have been able possible to use a local variable to manage whether the log has fired, this is a really robust way to handle it! In the future, this is a great place for a comment to explain the logic, but nicely done.

PR #1093, opened 2019-07-12
AIR-2389 (On Registration, "Log in" link takes users to the "link account" page)

jialin-he:
- `this.shibParameters` is an object, so check the email instead of checking itself


src/app/register/register.component.ts
```diff
@@ -197,7 +197,7 @@ export class RegisterComponent implements OnInit {
    * Navigates to correct page for login depending on whether or not shiboleth params exist
    */
   public navigateToLogin(): void {
-    if (this.shibParameters) {
+    if (this.shibParameters && this.shibParameters.email) {
```
bFraley: Good eye @jialin-he

PR #1095, opened 2019-07-17
AIR-2390

rehanabbasi:
AIW Banner: With these changes, we should be able to trigger a banner in AIW frontend by creating an incident in the statuspage.io dashboard.


rehanabbasi:
Make sure to get the `statuspage-client` installed through `yarn`

PR #1096, opened 2019-07-17
AIR-2416

rehanabbasi:
Group export: Poll for export status every 15 seconds if there is a timeout or a server error.


src/app/image-group-page/image-group-page.component.ts
```diff
@@ -428,11 +438,41 @@ export class ImageGroupPage implements OnInit, OnDestroy {
       })
       .catch( error => {
         console.error(error)
-        this.exportLoadingStateopts.state = LoadingState.error
```
bFraley: @rehanabbasi This is all looking really solid.

My only concern is that it appears that we'll never be setting `this.exportLoadingStateopts.state = LoadingState.error` or setting `this.exportLoadingStateopts.errorType = 'server'` again, which is assuming that all exports will never actually fail even after polling the check endpoint.

This is used in the "try again" logic.

src/app/image-group-page/image-group-page.component.ts
```diff
@@ -428,11 +438,41 @@ export class ImageGroupPage implements OnInit, OnDestroy {
...
+            this.downLoadFile(this.ig.name, downloadLink)
+          }, 5000)
+        }
+      }, (error) => {
+        console.error('Error in check status: ', error)
+      }
+    )
+  }
+
   public closeExportLoadingState(event?: any): void {
```
bFraley: @rehanabbasi I'm not sure of the best approach to take here. But, canceling the export now needs logic to clear the series of calls to the check endpoint.
```
clearInterval(this.exportStatusInterval)
clearInterval(this.loadingStateInterval)
```

rehanabbasi:
@bFraley the requested changes have been handled now, I believe this should be ready for review now!

PR #1101, opened 2019-07-18
AIR-2238

bFraley:
- Yarn add hammerjs

@codiak - I'm being optimistic here that just by adding hammerjs, our click events will be picked up from tap gestures. Otherwise, I'll add `tap` in the places related to this story.


corbfon:
😎 Hammer Time

PR #1109, opened 2019-07-26
AIR-2477: SearchV3: Clean out flag

codiak:


codiak:
@jialin-he Good catch, thanks Jialin

PR #1112, opened 2019-07-31
AIR-2481

rehanabbasi:
Add to an existing group: Make sure to query for editable groups for the logged-in user using `&level=created`, while fetching groups list.


codiak:
@rehanabbasi, If we should be swapping to using "created" for private groups, we should probably change this logic in multiple places?

But additionally, I know this is the solution that was suggested, but I don't understand how it would impact whether or not the option is enabled in the dropdown? Wouldn't this technically be more exclusive/specific?

PR #1116, opened 2019-08-05
Air-2489

bFraley:
- In compound objects, we need to also accept sequence_number "0" values


codiak:
@bFraley I made an update so this would use sort -- that way if sequenceNum aren't available, we'll still show a thumbnail. Tested against prod... any thoughts?

PR #1124, opened 2019-08-07
AIR-2436: Fix tabbing on the image group help window

eslawski:
The `We've made some updates to groups!` modal has been enhanced to ensure tabbing to its different elements remains scoped to only the modal (not leaking out to the rest of the page). The approach taken in this PR is similar to that of the the `pwd-reset-component`.

Other Changes:
- Pressing escape at anytime will close the modal and give the original link element focus again.


codiak:
Thanks for taking the extra time to match implementation on this one :)

PR #1129, opened 2019-08-09
AIR-2470: Prevent chat widget from displaying on wrong pages

eslawski:
The zendesk chat widget sometimes takes a second to load. This race condition can sometimes cause the widget to end up on a page it is not supposed to. Consider the steps to reproduce:

1. User visits login page (the widget should be displayed here)
2. We start fetching the zendesk chat widget script
3. The user quickly navigates away from the login page (assume it a page the widget should NOT appear)
4. The zendesk chat widget is finally ready to be displayed, so it happily displays even though it is for the wrong page.

**Solution:**
Leverage the zendesk chat api to determine when the widget is actually done loading. That way we can further determine if the current url is one it should be displaying on. Reference: https://api.zopim.com/files/meshim/widget/controllers/LiveChatAPI-js.html

Open to any other suggestions anyone may have!

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -198,9 +198,9 @@ export class AppComponent {
             this._script.loadScript('zendesk')
               .then( data => {
                 if (data['status'] !== "server_rendered") {
-                  zE(() => {
-                    $zopim(() => {
-                      $zopim.livechat.setOnConnected(() => {
+                  window['zE'](() => {
+                    window['$zopim'](() => {
+                      window['$zopim'].livechat.setOnConnected(() => {
                         // Sometimes the user navigates away from a page containing the chat widget
                         // but it hasn't loaded yet. Need to check once more after it loads to ensure
                         // we still should display it.
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -197,7 +197,6 @@ export class AppComponent {
           if (this.shouldShowChat(window.location.href) && this._app.config.showZendeskWidget) {
             this._script.loadScript('zendesk')
               .then( data => {
-                if (data['status'] !== "server_rendered") {
                   window['zE'](() => {
                     window['$zopim'](() => {
                       window['$zopim'].livechat.setOnConnected(() => {
@@ -212,7 +211,6 @@ export class AppComponent {
                       })
                     })
                   });
-                }
               })
               .catch( error => console.error(error) )
           } else {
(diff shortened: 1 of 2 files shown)


src/sass/app.scss
```diff
@@ -340,4 +340,9 @@ body body{
   & .arrow {
     left: calc(50% - 0.4rem);
   }
-}
\ No newline at end of file
+}
+
+// Hide Zendesk Chat until it is ready to be displayed
+.zopim {
```
eslawski: This makes it so the widget is hidden by default, until it is ready to be displayed. I found this necessary because the `setOnConnected` will display the widget for a brief moment before it figures out it needs to disappear.

src/app/app.component.ts
```diff
@@ -308,10 +330,10 @@ export class AppComponent {
   }
 
   // Show the chat widget on: 'login', 'browse/library', or 'browse/groups/public'
-  private showChatWidget(eventUrl: string): boolean {
+  private shouldShowChat(eventUrl: string): boolean {
       if (eventUrl.indexOf('browse/library') > -1 ||
           eventUrl.indexOf('browse/groups/public') > -1 ||
-          eventUrl.indexOf('login') > -1) {
+          eventUrl.indexOf('/login') > -1) {
```
eslawski: small bonus change. I don't think it was causing an issue here but if the string 'login' appeared anywhere in the url, then the widget would show. I think adding a `/` in front of `login` will help cut down the possibility unintended matches in the future.

src/sass/app.scss
eslawski: Had to abandon this change as there is some weirdness with if the user tries to dismiss the chat window. It's somewhat had to explain, but I can easily demo if anyone is interested.

So for now if the widget attempts to display erroneously on the page it might appear for a very brief moment before it dismisses itself.

src/app/app.component.ts
```diff
@@ -308,10 +330,10 @@ export class AppComponent {
   }
 
   // Show the chat widget on: 'login', 'browse/library', or 'browse/groups/public'
-  private showChatWidget(eventUrl: string): boolean {
+  private shouldShowChat(eventUrl: string): boolean {
```
bFraley: I like the more correct descriptive name here 👍

src/app/app.component.ts
```diff
@@ -235,6 +240,23 @@ export class AppComponent {
     }
   }
 
+  private hideZendeskChat() {
```
bFraley: @eslawski Good decision here to separate these out cleanly!

PR #1135, opened 2019-08-13
Air 2484: New implementation of creator/subject links with pipe

codiak:


src/app/shared/datatypes/asset.interface.ts
```diff
@@ -81,12 +81,14 @@ export class Asset {
             if (data.fieldName === 'License' && data.fieldValue === 'ADL Terms and Conditions') {
                 data.fieldValue = 'Use of this image is in accordance with the <a href="https://www.artstor.org/artstor-terms/" target="_blank">Artstor Terms & Conditions</a>'
             }
-            // Split by semicolons without removing semicolon
-            let values = data.fieldValue.split('|')
+            
+            let values = Array.isArray(data.fieldValue) ? data.fieldValue : [data.fieldValue]
```
eslawski: Can we no longer assume that `data.fieldValue` will be a string?

codiak: We should be able to, I had been experimenting with passing other values

codiak: Cleaned it out 👍

src/app/shared/filter-link.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+            restoreChar = ';'
+        } else {
+            // Pipes - handle splitting
+            values = value.split(/\|/g)
+        }
+        // Build search links wrapping field value
+        let linkHtml = ''
+        values.forEach((value, index) => {
+            if (index == (values.length - 1)) restoreChar = ''
+            linkHtml += `<a href="/#/search;${filterKey}:(${value})">${value}</a>${restoreChar}<span>&ensp;</span>`
```
eslawski: I noticed when you click one of these links it doesn't autopopulate the search text box. Should it? Comparing to PROD the only difference I notice is the `/` after `search`. Like so?

```suggestion
            linkHtml += `<a href="/#/search/${filterKey}:(${value})">${value}</a>${restoreChar}<span>&ensp;</span>`
```

PR #1140, opened 2019-08-15
AIR-2381: Fix Creative Common License icons/links

eslawski:
Since implementing the CC icons/links under AIR-2270, the controlled list of license names for Creative Commons have slightly changed in Forum. This has caused an issue on the front end since we are no longer matching on our expected string.

Here are some examples:
https://stage.artstor.org/#/asset/1002989035
https://stage.artstor.org/#/asset/1003307364

Here is an example of what we should expect the `License` metadata field to look like when it is for CC:
image

### DIFF
--- a/src/app/asset-page/licenses.ts
+++ b/src/app/asset-page/licenses.ts
@@ -27,7 +27,7 @@ export const licenses = [
       link: 'https://creativecommons.org/licenses/by-nc-nd/4.0/',
       img: 'https://licensebuttons.net/l/by-nc-nd/4.0/88x31.png'
     },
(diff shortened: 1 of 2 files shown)


src/app/asset-page/licenses.ts
```diff
@@ -3,36 +3,33 @@
  */
 
 export const licenses = [
-    { name: 'Attribution',
+    { name: 'Creative Commons: Attribution',
```
eslawski: The difference between the new and old values are mostly just the `Creative Commons:` bit at the beginning. Once exception is the `Public Domain Mark` at the bottom.

PR #1144, opened 2019-09-05
PLATO-126: Verify License field on PROD: Support "None" and "ADL" options

jialin-he:
- Boilerplate needs to be exist for all conditions, so remove the licenseText variable and hardcode the boilerplate into the template

### DIFF
--- a/src/app/asset-page/asset-page.component.ts
+++ b/src/app/asset-page/asset-page.component.ts
@@ -67,10 +67,6 @@ export class AssetPage implements OnInit, OnDestroy {
     public rightsLink: string = ''
     public rightsImg: string = ''
 
-    // License Statements values
-    public licenseLink: string = ''
-    public licenseImg: string = ''
-
     // Toast Variables
     public showToast: boolean = false
     public toastType: string = ''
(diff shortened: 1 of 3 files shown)


src/app/asset-page/asset-page.component.ts
```diff
@@ -864,12 +863,11 @@ export class AssetPage implements OnInit, OnDestroy {
         })
 
         if (licenseToUse) {
-          this.licenseText = licenseToUse.name
           this.licenseLink = licenseToUse.link
           this.licenseImg = licenseToUse.img
```
michael-iden: Something about the license mapping still seems a little fishy. I think it might be due to the nested loops between the find code and the outer loop [here](https://github.com/ithaka/aiw-ui/blob/94facfe95ccbc3af85296cf5166eeca0067fefd1/src/app/asset-page/asset-page.component.ts#L579-L583), and we save the result to an array but we also have some these component level fields that get overwritten. 
I don't know the right way to fix this at the moment. Maybe we need to return the image and link with the name, and then remove the component level link/image?

jialin-he: Now I attach the image and link to a license object, but still doing it in the component level though.

michael-iden (approved):

PR #1145, opened 2019-09-05
PLATO-165: Recently created groups not showing up in the save dialog box

jialin-he:
- Cannot reproduce on my machine
 - If looking further, the only place where we don't set `loading.recentGroups` back to false (this is why the three dots show) is when the item service and group service fails. So the problem is very likely to be cause by service failure.
 - Add an error message to notify users when the service fails

### DIFF
--- a/src/app/modals/add-to-group/add-to-group.component.ts
+++ b/src/app/modals/add-to-group/add-to-group.component.ts
@@ -51,6 +51,10 @@ export class AddToGroupModal implements OnInit, OnDestroy, AfterViewInit {
     recentGroups: false,
     allGroups: false
   }
+  public error: any = {
+    recentGroups: false,
+    allGroups: false
+  }
 
   public detailViewBounds: ImageZoomParams = {}
   public selectedGroup: any = {}
@@ -341,6 +345,7 @@ export class AddToGroupModal implements OnInit, OnDestroy, AfterViewInit {
           })
           .catch( error => {
             console.error(error)
+            this.error.recentGroups = true
             this.loading.recentGroups = false
           })
         } else {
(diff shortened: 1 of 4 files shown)


michael-iden (approved):
👍

src/app/modals/add-to-group/add-to-group.component.ts
```diff
@@ -341,6 +345,8 @@ export class AddToGroupModal implements OnInit, OnDestroy, AfterViewInit {
           })
           .catch( error => {
             console.error(error)
+            this.error.recentGroups = true
+            this.loading.recentGroups = false
```
eslawski: I looks like we have a lot of different situations where we set `this.loading.recentGroups = false`. Wondering if we could consolidate all of these into one place somehow. Seems like we want to stop loading recent groups regardless of the fact if it errors out or processes successfully. Maybe there is a way we can clean up this observable?

src/app/modals/new-ig-modal/new-ig-modal.component.ts
```diff
@@ -402,9 +402,9 @@ export class NewIgModal implements OnInit, AfterViewInit {
    * @param group the image group which needs the public property changed
    * @param isPublic the value to set the public property to
```
eslawski: I think we can remove this parameter now?

jialin-he: Oh yeah! Good eye!

src/app/modals/add-to-group/add-to-group.component.ts
eslawski: As we discussed offline looks like this might be a bigger endeavor than we thought. I created PLATO-192 for taking a closer look at this logic and learning more about observables

eslawski (approved):
Small nit, about removing that outdated parameter. Otherwise looks good!

michael-iden (approved):

PR #1146, opened 2019-09-10
PLATO-149: Add CC0 license

eslawski:
Resolves [PLATO-149](https://jira.jstor.org/browse/PLATO-149)

Adds the "Creative Commons: Free Reuse (CC0)" license

image

### DIFF
--- a/src/app/asset-page/licenses.ts
+++ b/src/app/asset-page/licenses.ts
@@ -8,7 +8,7 @@ export const licenses = [
       img: 'https://licensebuttons.net/l/by/4.0/88x31.png'
     },
     { name: 'Creative Commons: Free Reuse (CC0)',
-      link: 'https://creativecommons.org/share-your-work/public-domain/cc0/',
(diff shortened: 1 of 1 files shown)


src/app/asset-page/licenses.ts
```diff
@@ -7,6 +7,10 @@ export const licenses = [
       link: 'https://creativecommons.org/licenses/by/4.0/',
       img: 'https://licensebuttons.net/l/by/4.0/88x31.png'
     },
+    { name: 'Creative Commons: Free Reuse (CC0)',
```
bFraley: @eslawski reminder to be on the lookout for issues opened related to this license, as some publishers may add the license as something like, just "CCO", as we've seen that happen.

eslawski: 👍  Will do. My understanding is that these licenses are now actually a controlled list of options. If these happen to change I would think we would need to talk with the service teams.

PR #1156, opened 2019-10-07
PLATO-245: Images randomly moving when reordering images in groups

jialin-he:
Resolves [PLATO-245](https://jira.jstor.org/browse/PLATO-245)

test group on air01:
- `19949ef0-76fe-4fc2-bb60-bf622f3fd982`(without deleted images)
- `e10b0d18-6e42-41e1-992c-590502413f57` (with deleted images)


eslawski:
Change looks pretty straight forward, but not sure I fully understand it. Maybe a quick explanation in the PR description could provide some more context?

michael-iden (approved):
👍 Fix seems like it should work and matches what we do earlier in the file.

https://github.com/ithaka/aiw-ui/blob/0a91e5e442e7d4453fbd41d4c80052dc0c0aeaf3/src/app/asset-grid/asset-grid.component.pug#L115

michael-iden:
Finding the keyboard commands are now off with this change. Moving past a hidden item requires two arrow presses. I have an idea about making a `visibleResults` list of records that we can use for keyboard commands and the sortable/draggable grid

michael-iden (changes requested):
Removing approval for reason above

jialin-he:
Going to make a `visibleResults` list of records so that we don't need to hide unavailable items from the template.

src/app/asset-grid/asset-grid.component.ts
```diff
@@ -653,6 +654,11 @@ export class AssetGrid implements OnInit, OnDestroy {
           this.isLoading = false;
           this.allResults = allThumbnails
           this.results = this.allResults.slice(0)
+          this.visibleResults = []
+          for (let i = 0; i < this.results.length; i++) {
+              if (this.results[i].status !== 'not-available')
+                this.visibleResults.push(this.results[i])
```
michael-iden: nit - I think this this could be simplified slightly by using a filter on the results array. 

 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

jialin-he: Done!

src/app/asset-grid/asset-grid.component.ts
```diff
@@ -699,6 +705,7 @@ export class AssetGrid implements OnInit, OnDestroy {
 
   private saveReorder(): void {
     this.isLoading = true;
+    this.results = this.visibleResults
```
michael-iden: As mentioned offline, I think this will remove any items that are no longer available because they were deleted by the creator. 

Currently we give a warning message to the user with the option to cull those deleted items (see image below) but reordering will do that now automatically. Is there any situation where an item might not be "available" but the user would still want to keep it in their group? Does availability only change when an item is removed or could it be updated when institutional access is removed? 

image

src/app/asset-grid/asset-grid.component.ts
```diff
@@ -655,6 +656,7 @@ export class AssetGrid implements OnInit, OnDestroy {
           this.allResults = allThumbnails
           this.results = this.allResults.slice(0)
           this.visibleResults = this.results.filter(asset => asset.status !== 'not-available')
+          this.invisibleResults = this.results.filter(asset => asset.status === 'not-available')
```
michael-iden: I suggested `filter` when we just had one array, but now that we have two arrays it might make more sense to iterate over it just once. We could still use a `forEach` to stay with the control flow statements and inside the method conditionally push onto the visible/invisible results array.

jialin-he: Done!

michael-iden (approved):
LGTM!

eslawski (approved):
looks good!

bFraley (approved):
@michael-iden @jialin-he Nice improvement here.

PR #1159, opened 2019-10-15
PLATO-282: Use V2 endpoint to make artstor image groups public

jialin-he:
Change from `api/v1/group/{id}/admin/public` to `api/v2/group/{id}/admin/public`

### DIFF
--- a/src/app/_services/group.service.ts
+++ b/src/app/_services/group.service.ts
@@ -13,7 +13,6 @@ export class GroupService {
 
     private groupUrl: string
     private groupV1: string
-    private groupV2: string
     private options: {}
 
     public hasPrivateGroup: Observable<any>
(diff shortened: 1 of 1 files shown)


momo-sa (approved):

src/app/_services/group.service.ts
```diff
@@ -326,7 +328,7 @@ export class GroupService {
      * @returns Observable resolved with { success: boolean, message: string }
      */
     public updateIgPublic(igId: string) {
-        let reqUrl = [this.groupV1, igId, 'admin', 'public'].join('/')
+        let reqUrl = [this.groupV2, igId, 'admin', 'public'].join('/')
```
michael-iden: I think we can avoid creating a new `groupV2` var if we use `this.group`. 

https://github.com/ithaka/aiw-ui/blob/872076c34cd17a565aa4df3e79add69b9d0b1c30/src/app/_services/group.service.ts#L27

jialin-he: OH!!!!!!!!!!!!

jialin-he: Done!

michael-iden (approved):

PR #1161, opened 2019-10-17
PLATO-238: Constrained Zendesk chat widget to specific urls only

mrtoadsc:
Created an array of strings containing only the anchor/hash part of the url in the SPA. Checked the url hash fragment against the list and would show/load the chat widget by displaying/undisplaying the iFrame.

Small simplification and refactor of the solution.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -260,7 +260,7 @@ export class AppComponent {
   private showZendeskChat() {
     let zendDeskElement = this._dom.byId('launcher')
     if (zendDeskElement) {
-      zendDeskElement.style.display = null;
+      zendDeskElement.style.display = 'initial';
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -241,19 +251,16 @@ export class AppComponent {
...
   }
 
   private showZendeskChat() {
-    let zendeskElements = this._dom.bySelectorAll('.zopim')
-
-    if (zendeskElements && zendeskElements.length > 1) {
-      zendeskElements[0]['style']['display'] = 'block' // ChatWidgetButton
+    let zendDeskElement = this._dom.byId('launcher')
+    if (zendDeskElement) {
+      zendDeskElement.style.display = null;
```
michael-iden: The display null looks a little unusual, although it seems to work. Would `initial` be a more standard value for display?

mrtoadsc: A colleague of mine - who is a bit of a js expert - recommended this approach. It will reset itself to the original value of the dom. Sort if like a reset. 

I'll investigate the use of `initial` since the intent is much clearer

mrtoadsc: Tested with `initial` and it works the same so I've changed it from null to `initial`

PR #1172, opened 2019-10-29
PLATO-334: Remove ethnio survey

michael-iden:
[PLATO-334](https://jira.jstor.org/browse/PLATO-334)

Remove ethnio survey from the asset and image group pages

### DIFF
--- a/src/app/_services/script.store.ts
+++ b/src/app/_services/script.store.ts
@@ -6,6 +6,8 @@ interface Scripts {
 export const ScriptStore: Scripts[] = [
   // Zendesk chat widget
   { name: 'zendesk', src: '../../assets/js/zendesk.js' },
+  // Ethnio, current survey: Insights team on asset and image groups
+  // { name: 'ethnio-survey', src: '//ethn.io/45815.js' },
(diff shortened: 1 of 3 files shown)


src/app/_services/script.store.ts
```diff
@@ -6,8 +6,6 @@ interface Scripts {
 export const ScriptStore: Scripts[] = [
   // Zendesk chat widget
   { name: 'zendesk', src: '../../assets/js/zendesk.js' },
-  // Ethnio, current survey: Insights team on asset and image groups
-  { name: 'ethnio-survey', src: '//ethn.io/45815.js' },
```
michael-iden: I wasn't really sure if I should leave this here so that its maybe a little easier to add surveys? Or is the `45815` unique to a specific survey and we would have to change it anyways?

jialin-he: Usually we just comment it out instead of deleting it.

jialin-he: ...so that next time we only need to change the survey number.

michael-iden: 👍 Will fix

PR #1173, opened 2019-10-30
SRCHX-295: Fix large thumbnail view for multi view items

rehanabbasi:
Changes in this PR ensure that we are able to show large thumbnail views for multi view items in asset grid.


src/app/_services/thumbnail.service.ts
```diff
@@ -125,7 +125,8 @@ export class ThumbnailService {
     if (thumbnail.status != 'not-available' && thumbnail.isDetailView && !thumbnail.isDowngradedMedia) {
       return this.makeDetailViewThmb(thumbnail)
     } else if (thumbnail.isMultiView || (thumbnail.isDetailView && thumbnail.isDowngradedMedia)) {
-      return thumbnail.thumbnailImgUrl
+      let imgPath = thumbnail.size ? thumbnail.thumbnailImgUrl.replace(/(size)[0-4]/g, 'size' + thumbnail.size) : thumbnail.thumbnailImgUrl
```
NazimAli2017: What is this regex for? `/(size)[0-4]/g`

rehanabbasi: We are using this regex to identify the current thumbnail size substring (within the current thumbnail image url) and replace it with the required size set in the thumbnail object.

eslawski: Curious, what do thumbnail sizes of 1, 2, 3, or 4 represent? Where does that come from?

NazimAli2017: I think it would be a good idea to extract that regex to a constant and/or put the description either as the name or a comment about it and where the sizes come from. If someone reads this code, it won't be apparent what it is without context.

rehanabbasi: @eslawski the thumbnail sizes are part of the `thumbnailImgUrl: "/28e6a4ad-b586-4667-a3b3-267553770dc8_size1"` & `largeImgUrl: "/28e6a4ad-b586-4667-a3b3-267553770dc8_size2"`  strings in the group items endpoint https://stage.artstor.org/api/v1/group/60f18e33-94f4-4d29-a49d-aef48e55d254/items?object_ids=1003272527&object_ids=1003272515&object_ids=APRINCETONIG_10313684987&object_ids=AINDIANAIG_10313598345 response. These number numbers (size1, size2..) represent the thumbnail image size.

rehanabbasi: @NazimAli2017  I have made the suggested changes to make the updates more readable, in the recent commit `3c11d8e`

PR #1179, opened 2019-11-07
PALS-165( new artstor_item_view C5 fields)

bFraley:
- AssetPage: Adds new method `trackItemView` with logic for determining access values.
- Adds `reason_for_authentication` `has_access` and `entitlement_map` to additional fields.

*NOTE* - There is some cleanup unrelated to issue 165, removing needless commented out code.

### DIFF
--- a/src/app/asset-page/asset-page.component.ts
+++ b/src/app/asset-page/asset-page.component.ts
@@ -896,8 +896,7 @@ export class AssetPage implements OnInit, OnDestroy {
         item_id: asset.id,
         additional_fields: {
           has_access: hasAccess,
-          reason_for_authorization: reasonForAuth,
-          entitlement_map: {}
+          reason_for_authorization: reasonForAuth
         }
       })
     }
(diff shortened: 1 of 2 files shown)


src/app/asset-page/asset-page.component.ts
```diff
@@ -870,6 +866,42 @@ export class AssetPage implements OnInit, OnDestroy {
...
+    private trackItemView(asset: Asset): void {
+
+      let reasonForAuth: string[]
+      let hasAccess: boolean = true
+
+      if (this.showAccessDeniedModal) {
+        reasonForAuth = ['not_authorized']
+        hasAccess = false
+      }
+      else if (this._auth.isPublicOnly()) {
```
crystalfinch: Mind putting the `else if` on the same line as the closing bracket for the `if`?

bFraley: @crystalfinch Good eye, thanks! Updated.

PR #1181, opened 2019-11-13
PALS-144: Zooming and scaling must not be disabled

crystalfinch:
Disabled no more! ✨


crystalfinch:
ui deploy article-view

PR #1183, opened 2019-11-15
PLATO-364: Add logging on front end for downloading images

eslawski:
Resolves [PLATO-364](https://jira.jstor.org/browse/PLATO-364)

The `artstor_download_single_image` event is currently logged on backend `artstor-download-service` when user downloads image or detail zoom.

After discussing with STORK, we will be moving the logging to the front end in order to move forward with the C5 work. Once these changes go out, we have a story to remove the logging from the service (See [PLATO-366](https://github.com/ithaka/artstor-download-service/pull/121))

This PR adds front end logging for the following events and :

- `artstor_download_single_image`
- `artstor_image_download_view`

### DIFF
--- a/src/app/_services/image-group.service.ts
+++ b/src/app/_services/image-group.service.ts
@@ -82,7 +82,7 @@ export class ImageGroupService {
   }
 
   public triggerGoogleSlides(): void {
-    this.igDownloadTrigger.emit('GoogleSlides');
+    this.igDownloadTrigger.emit(SupportedExportTypes.GOOGLE_SLIDES);
   }
 
   /**
--- a/src/app/_services/image-group.service.ts
+++ b/src/app/_services/image-group.service.ts
@@ -17,6 +17,9 @@ import { ImageGroup, ImageGroupDescription } from './../shared'
  *
  */
 
+// The max number of images the user is able to bulk export.
+const MAX_DOWNLOAD_AMOUNT = 150;
+
 @Injectable()
 export class ImageGroupService {
   private proxyUrl = '';
@@ -83,8 +86,25 @@ export class ImageGroupService {
     this.igDownloadTrigger.emit('GoogleSlides');
   }
 
+  /**
+   * Given the image group determine the ids that are to be download given the user's access.
+   */
+  public getImageIdsToDownload(group: ImageGroup) {
+
+    /**
+     * We're using this as an auth check - instead of simply using all of the group.items strings and calling for downloads,
+     *  which was allowing restricted assets to be downloaded, we first ask for each assets' thumbnail, which will ensure
+     *  that only assets which the user has access to are returned
+     */
+
+    return this._assets.getAllThumbnails({ itemObjs: group.items }, group.id)
+      .then((thumbnails) => {
+        return thumbnails.slice(0, MAX_DOWNLOAD_AMOUNT).map((thumbnail) => thumbnail.id)
+      })
+  }
+
   /** Gets the link at which the resource can be downloaded. Will be set to the "accept" button's download property */
-  public getDownloadLink(group: ImageGroup, zip ?: boolean): Promise<any> {
+  public getDownloadLink(group: ImageGroup, imageIdsToDownload: string[], zip ?: boolean): Promise<any> {
     let header = new HttpHeaders().set('content-type', 'application/x-www-form-urlencoded')
     let options = { headers: header, withCredentials: true }
     let useLegacyMetadata: boolean = true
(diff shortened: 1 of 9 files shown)


src/app/_services/image-group.service.ts
```diff
@@ -17,6 +18,9 @@ import { ImageGroup, ImageGroupDescription } from './../shared'
  *
  */
 
+// The max number of images the user is able to bulk export.
+const MAX_DOWNLOAD_AMOUNT = 150;
```
eslawski: 150 is the max number of images that can be exported/downloaded at one time. This restriction was actually only enforced on the service side. For example, if an image group contained say 300 images, we on the front end would pass ALL 300 image ids to the service, but the service would only return 150 in the zip/pptx. 

Since we need to log only the 150 images actually downloaded, we needed to move this restriction to the front end.

src/app/_services/image-group.service.ts
```diff
@@ -72,71 +76,74 @@ export class ImageGroupService {
...
-
-    if (!zip) {
-      format = 'pptx'
-    } else {
-      format = 'zip'
-    }
+  /**
+   * Given the image group determine the ids that are to be download given the user's access.
+   */
+  public getImageIdsToDownload(group: ImageGroup) {
```
eslawski: Needed to extract this logic out into it's own function because we need to remember which images are actually being downloaded when we are ready to log them.

src/app/asset-page/asset-page.component.ts
```diff
@@ -866,40 +853,56 @@ export class AssetPage implements OnInit, OnDestroy {
...
+  /**
+   * trackEvent
+   * send captain's log events associated with asset
+   * @param asset
+   * @param eventType
+   * @returns void
+   */
+  private trackEvent(asset: Asset, eventType: string): void {
+    const hasAccess = !this.showAccessDeniedModal
+    const reasonForAuth = this.showAccessDeniedModal ? "not_authorized" : "license"
```
eslawski: I was able to simplify the `reasonForAuth` check. After discussing with auth team, all content on Artstor is managed via a license. Therefore even public/open images are technically managed by a license, and we can say the `reasonForAuth` is `license`.

src/app/image-group-page/image-group-page.component.ts
```diff
@@ -365,96 +361,6 @@ export class ImageGroupPage implements OnInit, OnDestroy {
     return totalTime/intervals
   }
 
-  private getPPT(): void{
```
eslawski: This function was nearly identical to `getZip`. Combined into one function called `executeBulkExport` a little bit farther below

src/app/_services/image-group.service.ts
momo-sa: 🤔 So if the service chances to 100, it'll cause a problem? Is it possible to make the service the source of truth?

src/app/image-group-page/image-group-page.component.ts
```diff
@@ -533,4 +443,56 @@ export class ImageGroupPage implements OnInit, OnDestroy {
...
+      exportType: exportType,
+      state: LoadingState.loading,
+      progress: 0,
+      imageIdsToDownload: imageIdsToDownload
+    }
+    this.showExportLoadingState = true
+
+    // Mimmic loading behaviour in intervals
+    this.loadingStateInterval = setInterval(() => {
+      this.exportLoadingStateopts.progress += 5
```
momo-sa: Should constants like this be defined?

src/app/_services/image-group.service.ts
```diff
@@ -72,71 +76,74 @@ export class ImageGroupService {
   }
 
   public triggerPPTExport(): void {
-    this.igDownloadTrigger.emit('ppt');
+    this.igDownloadTrigger.emit(SupportedExportTypes.PPTX);
   }
 
   public triggerZIPExport(): void {
-    this.igDownloadTrigger.emit('zip');
+    debugger;
```
michael-iden:
```suggestion
```

src/app/_services/image-group.service.ts
eslawski: For sure. TBH we are kinda of stuck right now. The `artstor-download-service` is currently logging this event. And it works because they know exactly which images are being downloaded. They have been pushing back on us to convert this log to the front end. But we really don't know for certain what actually gets downloaded. Maybe we should be getting a response from them that contains a complete list of all the ids. Or we can make an argument to continue logging on the service side.

src/app/asset-page/asset-page.component.pug
```diff
@@ -314,7 +314,13 @@
         ang-thumbnail.card.card--asset(*ngFor="let asset of prevAssetResults.thumbnails", [thumbnail]="asset", [ngClass]="{ 'selected' : asset.selected || isPrimaryAsset(asset) }", (click)="toggleAsset(asset)")
         .text-center.pt-2
           pagination([pageObj]="pagination", (goToPage)="goToPage($event)", style="display:inline-block")
-ang-agree-modal(*ngIf="showAgreeModal", (closeModal)="showAgreeModal = false", (acceptedTerms)="acceptedTerms = true", [downloadUrl]="downloadUrl", [downloadName]="downloadName", [isMSAgent]="isMSAgent", [setDownloadView]="setDownloadView")
+ang-agree-modal(*ngIf="showAgreeModal",
```
michael-iden: nit - maybe move this `*ngIf=` to a newline?

src/app/_services/image-group.service.ts
michael-iden: I think a conversation with them is warranted. The list of IDs downloaded is interesting. Talking offline, I also brought up the idea of returning a 400 response if the number of images to download is over the limit.

src/app/image-group-page/image-group-page.component.ts
```diff
@@ -324,33 +303,50 @@ export class ImageGroupPage implements OnInit, OnDestroy {
...
+            } else {
+              this.showGoogleAuth = true
+            }
+            break
+          }
+          case SupportedExportTypes.ZIP: {
+            // Perform ZIP download action
+            this.executeBulkExport(imageIdsToDownload, SupportedExportTypes.ZIP)
+            break
+          }
```
michael-iden: This case could be combined with PPTX and the second arg can become the switch condition. 
```
this.executeBulkExport(imageIdsToDownload, exportType)
```

src/app/image-group-page/image-group-page.component.ts
```diff
@@ -324,33 +306,44 @@ export class ImageGroupPage implements OnInit, OnDestroy {
...
-          console.log('Export to GS and show loading state')
-        } else {
-          this.showGoogleAuth = true
+    this._ig.getImageIdsToDownload(this.ig)
+      .then((imageIdsToDownload) => {
+        // If user has agreed, we should trigger download directly
+        switch (exportType) {
+          case SupportedExportTypes.PPTX:
+          case SupportedExportTypes.ZIP:
+            this.executeBulkExport(imageIdsToDownload, SupportedExportTypes.PPTX)
```
michael-iden:
```suggestion
            this.executeBulkExport(imageIdsToDownload, exportType)
```

eslawski: lol thanks.

src/app/_services/image-group.service.ts
eslawski: Meeting with STORK at 10am on Monday

eslawski: After talking with STORK they agreed the solution here is to continue logging this on the front end. I went ahead and removed the download limit and logging code from this PR. Check out this commit: https://github.com/ithaka/aiw-ui/pull/1183/commits/e92625af6ff09374706affbd68f5729916c22e76

src/app/asset-page/asset-page.component.ts
```diff
@@ -1787,6 +1790,46 @@ export class AssetPage implements OnInit, OnDestroy {
...
+    } else {
+      if(downloadType === "image") {
+        this.showAgreeModalForDownloadImage()
+      } else {
+        this.showAgreeModalForDownloadView()
+      }
+    }
+  }
+
+    public handleDownloadViewClick() {
```
michael-iden: Nit - Indenting here looks a little off, but it's hard to tell in this view

eslawski: Ya the indenting in this file seems messed up. Some places are 4 spaces, while other is 2. I messed with my editor settings trying to make it 4, but apparently we hardcode the indent size in this file: https://github.com/ithaka/aiw-ui/blob/f08ae76ec15fb700abbe4b8dd650985bdd8fcc15/.editorconfig#L9

So I went ahead and just fixed this file so that it is 2 everywhere.

PR #1186, opened 2019-11-20
PLATO-363: Fixes to c5 logs

eslawski:
- Fix error handling for exporting group
- Log asset id when access is not granted

### DIFF
--- a/src/app/asset-page/asset-page.component.ts
+++ b/src/app/asset-page/asset-page.component.ts
@@ -474,7 +474,7 @@ export class AssetPage implements OnInit, OnDestroy {
         // here is where we make the "access denied" modal appear
         if (!this.encryptedAccess) {
           this.showAccessDeniedModal = true
-          this.trackItemView(asset)
+          this.trackItemView(this.assetIds[0])
         } else {
           console.error('Failed to load externally shared asset', err)
           this.showServerErrorModal = true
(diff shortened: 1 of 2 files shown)


src/app/asset-page/asset-page.component.ts
```diff
@@ -906,20 +906,20 @@ export class AssetPage implements OnInit, OnDestroy {
...
     const hasAccess = !this.showAccessDeniedModal
     const reasonForAuth = this.showAccessDeniedModal ? "not_authorized" : "license"
-    const abSegments = this._search.ab_segments.get(asset.id)
+    const abSegments = this._search.ab_segments.get(assetId)
 
     this._log.log({
       eventType: eventType,
       referring_requestid: this._search.latestSearchRequestId,
       ab_segments: abSegments ? [abSegments] : [],
-      item_id: asset.id,
```
eslawski: When the user does not have access to the asset, then `asset` will be undefined (because it gets loaded from the metadata service). Converted over to using the `this.assetIds` because we should always have that info regardless of access or not.

src/app/image-group-page/image-group-page.component.ts
```diff
@@ -309,28 +309,24 @@ export class ImageGroupPage implements OnInit, OnDestroy {
   private exportGroup(exportType: SupportedExportTypes): void {
     this.showTermsConditions = false
 
-    this._ig.getImageIdsToDownload(this.ig)
```
eslawski: Moved the fetching of the `imageIdsToDownload` into the function that handle the loading states for the export. That is how we are able to reintroduce the error handling that used to be in place.

PR #1188, opened 2019-11-22
PLATO-391: Log doi for item specific events

eslawski:
Start logging the doi for the following event types

`artstor_view_item`
`artstor_download_single_image`
`artstor_image_view_download`


src/app/asset-page/asset-page.component.ts
```diff
@@ -922,20 +923,20 @@ export class AssetPage implements OnInit, OnDestroy {
       item_id: assetId,
       additional_fields: {
         has_access: hasAccess,
-        reason_for_authorization: [reasonForAuth]
+        reason_for_authorization: [reasonForAuth],
+        ...doi && {doi: doi}
```
crystalfinch: This line only add the doi if it's not null?

eslawski: That is correct!

PR #1189, opened 2019-11-25
PLATO-391: Pass doi as array

eslawski:
The artstor-log-service expects an array of dois, also moved it as a top level key after coordinating with blueprint team

### DIFF
--- a/src/app/asset-page/asset-page.component.ts
+++ b/src/app/asset-page/asset-page.component.ts
@@ -924,10 +924,10 @@ export class AssetPage implements OnInit, OnDestroy {
       referring_requestid: this._search.latestSearchRequestId,
       ab_segments: abSegments ? [abSegments] : [],
       item_id: assetId,
+      ...doi && {doi: [doi]},
       additional_fields: {
         has_access: hasAccess,
         reason_for_authorization: [reasonForAuth],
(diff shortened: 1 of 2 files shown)


src/app/asset-page/asset-page.component.ts
```diff
@@ -924,10 +924,10 @@ export class AssetPage implements OnInit, OnDestroy {
       referring_requestid: this._search.latestSearchRequestId,
       ab_segments: abSegments ? [abSegments] : [],
       item_id: assetId,
+      ...doi && {doi: [doi]},
```
michael-iden: What is the purpose of the `...`? Are we trying to do a spread?

eslawski: It's actually being used to conditionally add the `doi` key to the top level dictionary. Learned here: https://medium.com/@mikeh91/conditionally-adding-keys-to-javascript-objects-using-spread-operators-and-short-circuit-evaluation-acf157488ede

Let me know if that is overly confusing though. The alternative would probably be logging an empty array or `null`?

michael-iden: Ahh gotcha. Makes sense now. 👍

PR #1190, opened 2019-12-02
PLATO-384: removed ethnio survey

gbassham:
Removed the ethnio survey from asset, image groups, and search pages.


jialin-he:
I remember it is also on the asset page? Maybe want to remove it from that too

gbassham:
> I remember it is also on the asset page? Maybe want to remove it from that too

I removed it from three files, including asset-page.component.ts. Is there a different page you're thinking of? I can double-check and make sure it's not there.

PR #1191, opened 2019-12-02
SRCHX-138: Allow only one date filter on advanced search

willschoener:
Allows users to filter in advanced search by only 1 date filter, whether start or end.

Implements same functionality as side bar in browse page.

### DIFF
--- a/src/app/_services/assets.service.ts
+++ b/src/app/_services/assets.service.ts
@@ -701,7 +701,7 @@ export class AssetService {
                 }
             });
 
-            if (params['startDate'] && params['endDate']){
+            if (params['startDate'] || params['endDate']){
                 dateObj = {
                     modified : true,
                     earliest : {
--- a/src/app/_services/assets.service.ts
+++ b/src/app/_services/assets.service.ts
@@ -705,12 +705,12 @@ export class AssetService {
                 dateObj = {
                     modified : true,
                     earliest : {
-                        date : Math.abs(params['startDate']),
-                        era : params['startDate'] < 0 ? 'BCE' : 'CE'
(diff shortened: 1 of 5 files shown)


src/app/_services/assets.service.ts
```diff
@@ -701,7 +701,7 @@ export class AssetService {
                 }
             });
 
-            if (params['startDate'] && params['endDate']){
+            if (params['startDate'] || params['endDate']){
```
momo-sa: IMO the code below becomes a little obscure if we're not expecting integer values. In particular, we do math operations on undefined values:

```
                    earliest : {
                        date : Math.abs(params['startDate']),
                        era : params['startDate'] < 0 ? 'BCE' : 'CE'
                    },
```

Is there a way to clarify how we're handling the `undefined` case?

willschoener: With javascript, undefined becomes 0... Which is weird. This ticket was to have advanced search act the same way the side-facet bar on the browse page works. which defaults to year 0 for any year not entered. We have a ticket SRCHX-406 which will change this behavior

momo-sa: Oh, yeah, I don't doubt that it works. I just thought it looked obscure because it implicitly coerces `undefined`.

willschoener:
Putting back in progress to fix the date range issue when only one date filter is entered.

src/app/_services/assets.service.ts
momo-sa: For example, the following looks equivalent, but gives a different result because of the hidden case:

```
era : params['startDate'] >= 0 ? 'CE' : 'BCE'
```

PR #1196, opened 2019-12-17
PLATO-423: Log the current url for asset related actions

eslawski:
While recently debugging some artstor logging issues, we came across some interesting url parameters that affect can affect the way access is evaluated when viewing/downloading an asset. It would have been super helpful to have seen this url reflected in the logs in order to help draw conclusions. 

This PR just adds the full url to the asset related logging (viewing/downloading). Let me know if there are any concerns with this!

### DIFF
--- a/src/app/asset-page/asset-page.component.ts
+++ b/src/app/asset-page/asset-page.component.ts
@@ -911,10 +911,10 @@ export class AssetPage implements OnInit, OnDestroy {
   private trackEvent(eventType: string): void {
     const hasAccess = !this.showAccessDeniedModal
     const reasonForAuth = this.showAccessDeniedModal ? "not_authorized" : "license"
-    const assetId = this.assetIds[0]
-    const abSegments = this._search.ab_segments.get(assetId)
     const authorizedAsset = this.assets[0]
+    const assetId = authorizedAsset ? authorizedAsset.id : this.assetIds[0]
     const doi = authorizedAsset ? authorizedAsset.doi : null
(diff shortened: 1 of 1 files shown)


src/app/asset-page/asset-page.component.ts
```diff
@@ -911,10 +911,10 @@ export class AssetPage implements OnInit, OnDestroy {
   private trackEvent(eventType: string): void {
     const hasAccess = !this.showAccessDeniedModal
     const reasonForAuth = this.showAccessDeniedModal ? "not_authorized" : "license"
-    const assetId = this.assetIds[0]
-    const abSegments = this._search.ab_segments.get(assetId)
     const authorizedAsset = this.assets[0]
+    const assetId = authorizedAsset ? authorizedAsset.id : this.assetIds[0]
```
eslawski: Additional change while investigating PLATO-423:

So the `assetIds` array is populated directly from the id found in the url. I learned that the id present in the url does not always reflect the actual id of the image. For example: https://stage.artstor.org/#/asset/external/8DtZYyMmJloyLyw5ej1%252FR34o

in this special situation, this "encrypted" asset id eventually resolves to an actual image. And that is the id we should be logging. So pulling the id off the `asset` object (i.e what we get from the metadata service) is the correct one to be logging.

Still need a fallback on the `assetIds[0]` though because if the user is not authorized to view the asset, the metadata will not have loaded, hence `authorizedAsset` will be null.

PR #1197, opened 2019-12-19
SRCHX-392: Geographic filters in advanced search is using efq values

mrtoadsc:
Geographic queries will use the efq (encoded filter query) values from the search service. These values will be selected in the search results page.

The `SearchModal.loadAppliedFiltersFromURL()` method will parse the geography portion of the url to select the proper tree nodes.

This work has not been comprehensively tested. I only have one day left on this project (contract expires) and there are a lot of devs out of the office now so you folks will need to take this over the finish line.

### DIFF
--- a/src/app/modals/search-modal/search-modal.component.ts
+++ b/src/app/modals/search-modal/search-modal.component.ts
@@ -215,7 +215,46 @@ export class SearchModal implements OnInit, AfterViewInit {
     let filterParams = this.queryUtil.generateFilters(this.filterSelections, this.advanceSearchDate)
     let currentParams = this.route.snapshot.params
     let queryParams = {}
+
     // Consolidate filters with multiple applied
+    filterParams = this.combineMulitpleAppliedFilters(filterParams)
+
+    // Maintain feature flags
+    if (currentParams['featureFlag']) {
+      queryParams['featureFlag'] = currentParams['featureFlag']
+    }
+    
+    // Build Solr query string from filters
+    let orQuery: string = this.buildSolrQuery(advQuery, filterParams)
+
+    // Apply date filter
+    if(filterParams["startDate"]) {
+      queryParams["startDate"] = filterParams["startDate"]
+    }
+    if(filterParams["endDate"]) {
+      queryParams["endDate"] = filterParams["endDate"]
+    }
+
+    // Apply geography filter
+    if (filterParams['geography']) {
+      queryParams['geography'] = []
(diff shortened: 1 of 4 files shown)


src/app/modals/search-modal/search-modal.component.ts
```diff
@@ -233,14 +233,14 @@ export class SearchModal implements OnInit, AfterViewInit {
     // Construct OR query between filters within the same filter group
     // filters across multiple filter groups will be AND-ed
     // example orQuery: paints AND artclassification_str:"Photographs" OR artclassification_str:"Paintings" OR artclassification_str:"Prints" OR artclassification_str:"photographs" AND year:[-4000 TO 1980]
-    let orQuery: string = ''
+    let orQuery: string = advQuery
     // Do not apply "* AND", Solr will treat the wildcard as a character in that case
-    if (advQuery !== '*') {
-      orQuery = advQuery
-    }
+    // if (advQuery !== '*') {
```
NazimAli2017: Looks like this can be deleted.

src/app/modals/search-modal/search-modal.component.ts
```diff
@@ -266,6 +266,12 @@ export class SearchModal implements OnInit, AfterViewInit {
     if(filterParams["endDate"]) {
       queryParams["endDate"] = filterParams["endDate"]
     }
+    if (filterParams['geography']) {
```
NazimAli2017: Could be extracted to a smaller method and be unit tested, this function is already bloated.

src/app/modals/search-modal/search-modal.component.ts
```diff
@@ -295,6 +301,11 @@ export class SearchModal implements OnInit, AfterViewInit {
       this.loadingFilters = false
       return
     }
+    let geography: string = routeParams['geography']
+    if (geography) {
+      appliedFiltersObj['geography'] = geography.startsWith('[') ? JSON.parse(geography) : geography.split(',')
```
NazimAli2017: Same for this.

src/app/modals/search-modal/search-modal.component.ts
```diff
@@ -389,6 +400,42 @@ export class SearchModal implements OnInit, AfterViewInit {
             }
           }
           updateSelectedFilters = true
+          break
+        }
+
+        case 'geography': {
```
NazimAli2017: This case is extremely large, should be broken up into smaller methods.

PR #1198, opened 2019-12-20
PLATO-463: Artstor logging adjustments for view/download item events

eslawski:
Resolves PLATO-463

Changes:

- Data platform requested that we be a little more granular with our `reason_for_authorization` field. This will help for account for the edge cases that allow a user to view an image even though they don't technically have the entitlements to do so.

- We learned that the asset page is server side rendered if it is hit with the `/public` or `/object` urls. https://github.com/ithaka/aiw-ui/blob/884259db16fc0eb967f7ac819e81286c5c4cc078/server/server.ts#L101 This running node server was actually emitting a captains log in an attempt to pre-render the page. This log was incomplete and overall not C5 worthy. A log would then again be fired on the front end when the javascript would run in the actual browser. The log generated at this stage is well formed. So here we are just omitting the ssr log.

### DIFF
--- a/src/app/asset-page/asset-page.component.ts
+++ b/src/app/asset-page/asset-page.component.ts
@@ -902,6 +902,20 @@ export class AssetPage implements OnInit, OnDestroy {
     return false
   }
 
+  private getReasonForAuth() {
+    if(this.showAccessDeniedModal) {
+      return "not_authorized"
+    } else {
+      if(this.assetGroupId) {
+        return "image_group"
+      } else if(this.encryptedAccess) {
+        return "encrypted_access"
+      } else {
+        return "license"
+      }
+    }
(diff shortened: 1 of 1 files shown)


src/app/asset-page/asset-page.component.ts
```diff
@@ -902,6 +902,20 @@ export class AssetPage implements OnInit, OnDestroy {
     return false
   }
 
+  private getReasonForAuth() {
+    if(this.showAccessDeniedModal) {
+      return "not_authorized"
+    } else {
+      if(this.assetGroupId) {
+        return "image_group"
```
eslawski: The `image_group` and `encrypted_access` strings are just placeholders for now until we can chat with data platform.

PR #1199, opened 2019-12-26
PLATO-217-220: Add first-level headings to item view and image group pages

gbassham:
Add `h1` tags to item view and image group pages for accessibility. Part of VPAT Conformance.


src/app/asset-page/asset-page.component.pug
```diff
@@ -19,6 +19,7 @@
               span#assetPageCount.font-weight-bold {{ totalAssetCount - restrictedAssetsCount }}
               i.icon.icon-direction.icon-right(title="Next in results", (click)="showNextAsset()", (keyup.enter)="showNextAsset()", [class.disabled]="assetNumber >= totalAssetCount - restrictedAssetsCount", [attr.tabindex]="assetNumber >= totalAssetCount - restrictedAssetsCount ? -1 : 1", aria-label="Next item, press enter to go to next result item. You can also use left and right arrow keys to navigate between result items.", role="button")
     //- for accessibility so that page has title
+    h1.sr-only([innerHTML]="getTitleAndResults()")
     h1.sr-only(*ngIf="isFullscreen") Full screen mode
```
jialin-he: Would it be better if we combine those two h1s? Looks a little weird to me to have two screen-reader-only h1s on full screen mode.

gbassham: I do not have an issue with combining them, but I wonder whether the full screen mode should even be an h1? It doesn't make sense to me as a first-level heading.

eslawski:
Gayla and I discussed offline what implications there might be to SEO here by introducing an h1 on this page. We determined there will be no impact to current behavior as we are setting the page title directly and the `h1` being displayed is for screen readers only.

PR #1200, opened 2019-12-27
PLATO-221: Image Viewer Controls Keyboard Access

jialin-he:
Resolve PLATO-221 and PLATO-225

- Make all interactive elements accessible by keyboard in the viewer
- Prevent other interactive elements outside of the viewer from reachable while it is in full view mode

### DIFF
--- a/src/app/_services/toolbox.service.ts
+++ b/src/app/_services/toolbox.service.ts
@@ -83,9 +83,9 @@ export class ToolboxService {
      * @note Should fire immediately within a user-event binding (click, key)
      */
     public requestFullScreen(): void {
-        let el = document.body
+        let el = document.getElementById("artstor-viewer");
         // Supports most browsers and their versions.
-        var requestMethod = el.requestFullscreen || el['webkitRequestFullscreen'] || el['mozRequestFullscreen'] || el['msRequestFullscreen']
+        var requestMethod = el.requestFullscreen || el['webkitRequestFullscreen'] || el['mozRequestFullscreen'] || el['msRequestFullscreen'];
         if (requestMethod) { // Native full screen.
             requestMethod.call(el);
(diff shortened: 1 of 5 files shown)


src/app/asset-page/asset-page.component.pug
```diff
@@ -18,73 +18,109 @@
...
-            (removeAsset)="toggleAsset($event, i)",
-            (nextPage)="showNextAsset()",
-            (prevPage)="showPrevAsset()",
-            (multiViewPageViaArrow)="multiViewPageViaArrow()",
-            (multiViewPageViaThumbnail)="multiViewPageViaThumbnail()",
-            (assetDrawer)="toggleAssetDrawer(!showAssetDrawer)",
-            [legacyFlag]="false",
-            [openLibraryFlag]="fromOpenLibrary"
+
+    .col-md-8#artstor-viewer
```
jialin-he: Wrap all elements that are going to be shown in full screen mode in one div, so that I can only call request full screen on this div instead of the whole body. This is to prevent other interactive elements that are not showing in full screen mode from being accessible.

gbassham (approved):

src/app/_services/toolbox.service.ts
```diff
@@ -94,6 +94,8 @@ export class ToolboxService {
                 wscript.SendKeys("{F11}");
             }
         }
+        let htmlelement: HTMLElement = <HTMLElement>document.getElementsByClassName('btn--zoomIn')[0];
```
eslawski: nit: Could use a more specific name, like `zoomInButtonElement`?

jialin-he: Done!

eslawski (approved):

PR #1201, opened 2019-12-27
PLATO-223: Image Groups Modals Have Keyboard Trap

jialin-he:
Resolve PLATO-223

- Prevent keyboard trap for "Generate image link" and "Generate group link" modals
- Also realize same behavior on other modals

### DIFF
--- a/src/app/asset-page/agree-modal/agree-modal.component.ts
+++ b/src/app/asset-page/agree-modal/agree-modal.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, Output, Input, EventEmitter, ElementRef, ViewChild } from '@angular/core';
+import { Component, AfterViewInit, Output, Input, EventEmitter, ElementRef, ViewChild } from '@angular/core';
 
 // Project Dependencies
 import { AuthService, DomUtilityService } from '_services'
@@ -9,7 +9,7 @@ import { Asset } from './../asset'
   templateUrl: 'agree-modal.component.pug',
   styleUrls: [ './agree-modal.component.scss' ]
 })
-export class AgreeModalComponent implements OnInit {
+export class AgreeModalComponent implements AfterViewInit {
 
   /** Causes modal to be hidden */
   @Output()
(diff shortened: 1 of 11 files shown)


src/app/asset-page/agree-modal/agree-modal.component.ts
```diff
@@ -30,20 +30,21 @@ export class AgreeModalComponent implements OnInit {
...
-    // TO-DO: Only reference document client-side
-    // let htmlelement: HTMLElement = <HTMLElement>this._dom.byId('modal');
-    // htmlelement.focus()
-    if (this.modalElement && this.modalElement.nativeElement){
-      this.modalElement.nativeElement.focus()
+  ngAfterViewInit() {
+    this.startModalFocus()
+  }
+
+  // Set initial focus on the modal Title h4
```
eslawski: Curious if we want to be setting the focus on the h4. Traditionally I think we set focus to the first "interactive" element on the modal. Like in this case I think it would be the terms and conditions link?

jialin-he: There used to be the expectation to focus first on the heading so that the screen reader will read out what this modal is. I don't know if we want to change that.

eslawski: Ah I see. If that is the old expected behavior I say we just stick with that

src/app/modals/generate-citation/generate-citation.component.ts
```diff
@@ -56,8 +51,12 @@ export class GenerateCitation implements OnInit, AfterViewInit {
 
   // Set initial focus on the modal Title h1
   public startModalFocus() {
-    let modalStartFocus: HTMLElement = <HTMLElement>this._dom.byId('generate-citation-title')
-    modalStartFocus.focus()
+    // let modalStartFocus: HTMLElement = <HTMLElement>this._dom.byId('generate-citation-title')
```
eslawski: nit: can remove commented out code

src/app/image-group-page/terms-and-conditions/terms-and-conditions.component.pug
```diff
@@ -2,7 +2,7 @@
   .modal-dialog(role="dialog", (keydown.esc)="hideModal($event)")
     .modal-content
       .modal-header
-        h4.modal-title#term-condition-title(tabindex="0", (keydown.shift.tab)="hideModal($event)") {{ 'IMAGE_GROUP_PAGE.TERMS_AND_CONDITIONS.TITLE' | translate }}
+        h4.modal-title(#termConditionTitle, tabindex="0", (keydown.shift.tab)="hideModal($event)") {{ 'IMAGE_GROUP_PAGE.TERMS_AND_CONDITIONS.TITLE' | translate }}
```
eslawski: nit: `termsAndConditionsTitle`. Might also need to change in the component ts file

eslawski (approved):

gbassham (approved):

PR #1207, opened 2020-01-03
PLATO-221-225: Image Viewer Controls Keyboard Access

jialin-he:
- Prevent elements that are invisible in the viewer from getting focus
    - skip to main content link
    - nav menu
    -  footer
    -  thumbnails in the drawer when the drawer is not open
 - Set visible blue border when element gets focused
 - Hide pagination button for first and last page
 - Manually add event listeners to thumbnails in the reference strip to allow `page` event when hitting enter key
 - Set up correct tab order in the viewer (for fullscreen)
    1. Main canvas
    2. Zoom in, zoom out, fit to view
    3. Compare, exit
    4. Page arrows
    5. Thumbnails in the reference strip if it is multi-view
    6. Pagination buttons
    7. Hide caption, quiz mode toggle button

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -37,7 +37,7 @@ const STATUS_PAGE_CMP_ID_PROD: string = 'cmy3vpk5tq18'
         <button id="button" (click)="findMainContent()" (keyup.enter)="findMainContent()" tabindex="1" class="sr-only sr-only-focusable"> Skip to main content </button>
       </div>
       <nav-bar tabindex="-1"></nav-bar>
-      <main>
+      <main id="main">
         <router-outlet></router-outlet>
       </main>
 
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -304,24 +304,12 @@ export class AppComponent {
 
   private removeSkipToMainContent() {
     let skipToMainContentButton = document.getElementById("skip-main-content-button")
-    skipToMainContentButton.remove()
+    skipToMainContentButton.tabIndex = -1
   }
 
   private addSkipToMainContent() {
-    let skipToMainContentDiv = document.getElementById("skip-main-content-div")
-    // <button id="skip-main-content-button" (click)="findMainContent()" (keyup.enter)="findMainContent()" tabindex="1" class="sr-only sr-only-focusable"> Skip to main content </button>
-    let skipToMainContentButton = document.createElement("button")
-    skipToMainContentButton.innerHTML = "Skip to main content"
+    let skipToMainContentButton = document.getElementById("skip-main-content-button")
     skipToMainContentButton.tabIndex = 1
-    skipToMainContentButton.id = "skip-main-content-button"
-    skipToMainContentButton.setAttribute("class", "sr-only sr-only-focusable")
-    skipToMainContentButton.onclick = this.findMainContent
-    skipToMainContentButton.onkeyup = (event) => {
-      if (event.keyCode === 13) {
-        this.findMainContent()
-      }
-    }
-    skipToMainContentDiv.appendChild(skipToMainContentButton);
   }
 
   private hideZendeskChat() {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -33,8 +33,8 @@ const STATUS_PAGE_CMP_ID_PROD: string = 'cmy3vpk5tq18'
   template: `
     <ang-sky-banner *ngIf="showSkyBanner" [textValue]="skyBannerCopy" (closeBanner)="closeBanner()"></ang-sky-banner>
     <div>
-      <div id="skip" tabindex="-1">
-        <button id="button" (click)="findMainContent()" (keyup.enter)="findMainContent()" tabindex="1" class="sr-only sr-only-focusable"> Skip to main content </button>
+      <div id="skip-main-content-div" tabindex="-1">
+        <button id="skip-main-content-button" (click)="findMainContent()" (keyup.enter)="findMainContent()" tabindex="1" class="sr-only sr-only-focusable"> Skip to main content </button>
       </div>
       <nav-bar tabindex="-1"></nav-bar>
       <main id="main">
@@ -52,6 +52,8 @@ export class AppComponent {
   public showSkyBanner: boolean = false
   public skyBannerCopy: string = ''
   public test: any = {}
+
+  private isFullscreen: boolean = false
 
   public statusPageClient: any
   /**
@@ -141,7 +143,7 @@ export class AppComponent {
 
         if (isPlatformBrowser(this.platformId)) {
           // focus on the wrapper of the "skip to main content link" everytime new page is loaded
-          let mainEl = <HTMLElement>(this._dom.byId('skip'))
+          let mainEl = <HTMLElement>(this._dom.byId('skip-main-content-div'))
           if (!(event.url.indexOf('browse') > -1) && !(event.url.indexOf('search') > -1) && !(event.url.indexOf('asset') > -1)) // Don't set focus to skip to main content on browse pages so that we can easily go between browse levels
             mainEl.focus()
         }
@@ -273,6 +275,53 @@ export class AppComponent {
     }
 
     this.initPerimeterX();
+
+    document.addEventListener("fullscreenchange", () => {
+      this.toggleSkipToMainContent()
+    }, false);
+  
+    document.addEventListener("mozfullscreenchange", () => {
+      this.toggleSkipToMainContent()
+    }, false);
+  
+    document.addEventListener("webkitfullscreenchange",() => {
+      this.toggleSkipToMainContent()
+    }, false);
+  
+    document.addEventListener("msfullscreenchange", () => {
+      this.toggleSkipToMainContent()
+    }, false);
+  }
+
+  private toggleSkipToMainContent() {
+    if (this.isFullscreen) {
+        this.addSkipToMainContent()
+    } else {
+      this.removeSkipToMainContent()
+    }
+    this.isFullscreen = !this.isFullscreen
+  }
+
+  private removeSkipToMainContent() {
+    let skipToMainContentButton = document.getElementById("skip-main-content-button")
+    skipToMainContentButton.remove()
+  }
+
+  private addSkipToMainContent() {
+    let skipToMainContentDiv = document.getElementById("skip-main-content-div")
+    // <button id="skip-main-content-button" (click)="findMainContent()" (keyup.enter)="findMainContent()" tabindex="1" class="sr-only sr-only-focusable"> Skip to main content </button>
+    let skipToMainContentButton = document.createElement("button")
+    skipToMainContentButton.innerHTML = "Skip to main content"
+    skipToMainContentButton.tabIndex = 1
+    skipToMainContentButton.id = "skip-main-content-button"
+    skipToMainContentButton.setAttribute("class", "sr-only sr-only-focusable")
+    skipToMainContentButton.onclick = this.findMainContent
+    skipToMainContentButton.onkeyup = (event) => {
+      if (event.keyCode === 13) {
+        this.findMainContent()
+      }
+    }
+    skipToMainContentDiv.appendChild(skipToMainContentButton);
   }
 
   private hideZendeskChat() {
--- a/src/app/asset-grid/thumbnail/thumbnail.component.ts
+++ b/src/app/asset-grid/thumbnail/thumbnail.component.ts
@@ -44,6 +44,9 @@ export class ThumbnailComponent implements OnInit, OnChanges {
   @Input()
   public editMode: boolean
 
+  @Input()
+  private compareMode: boolean
+
   public src: SafeUrl
   // Keeps the track of multiViewItems count associated with the current asset
   public multiviewItemCount: number = 0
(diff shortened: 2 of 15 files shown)


src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
```diff
@@ -406,6 +406,29 @@ export class ArtstorViewerComponent implements OnInit, OnDestroy {
             this.osdViewer.destroy();
         });
 
+        this.osdViewer.addOnceHandler("tile-drawn", () => {
+            // Get all canvas elements, first one is the main canvas, second one is the navigator
```
eslawski: I think we omit the navigator when in full screen mode. As a result, our `canvasElements` might not always match the ordering we are expecting them to be.

Overall I am a little concerned with how fragile this tab logic can be. I don't have any great ideas at the moment, but I would love to brainstorm some with you!

jialin-he: Oh yes we omit the navigator in full screen mode. But this event is fired only once when the tiles are drawn, which is when we first land on this asset page. So by the time we open fullscreen, the event listeners have already been attached to those elements.

gbassham: Just off the top of my head, is it possible to give the elements that need the listeners a different class so you can look for that and don't have to rely on `openseadragon-canvas`?

eslawski: Hmm wonder if we will encounter any issues if the user opens up the viewer directly from an image group

jialin-he: @eslawski I just had another look, so in fullscreen mode, although we don't show the navigator, it still exists in the dom, so it won't mess up the order. I would be up for a better solution though.

jialin-he: @gbassham The problem is (I think), those elements are generated by openseadragon, I don't have control over them.

gbassham: Yeah, I was afraid you were going to say that.

src/app/app.component.ts
```diff
@@ -273,6 +275,41 @@ export class AppComponent {
...
+  }
+
+  private removeSkipToMainContent() {
+    let skipToMainContentButton = document.getElementById("skip-main-content-button")
+    skipToMainContentButton.tabIndex = -1
+  }
+
+  private addSkipToMainContent() {
+    let skipToMainContentButton = document.getElementById("skip-main-content-button")
+    skipToMainContentButton.tabIndex = 1
```
eslawski: So this adding/removing the "skip to main content" control... Is this still necessary now that we have changed the element we give to the `requestFullScreen`? If I understand correctly the `skip-main-content-button` lives outside of the `main` element.

jialin-he: Not necessary for chrome where the `requestFullScreen` works well with the element, but this doesn't work for Firefox. The dom still contains all body with Firefox.

eslawski: hmm interesting. I wonder why that is the case for firefox. Seems like we could save ourselves a lot of code if we didn't have to account for this weird edge case (I know we hide/show the nav/footer in this way too for that reason). It might be worth spending some more time trying to get to the bottom of that? This isn't happening on the jstor side of things right? I wonder if it has to do with the us feeding the full screen a `main` element as opposed to a div or something?

jialin-he: I tired div before I switched to `main`. It didn't work so it's not the problem of `main`.

michael-iden: Hmm interesting. I now remember that Openseadragon removes everything else from the DOM when it goes into full screen mode using the default controls, which creates an entirely different set of complexity. https://openseadragon.github.io/

Emulating this may be worth discussing but it may be pretty far out of scope.

src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
```diff
@@ -69,6 +69,11 @@ $desktop: 992px;*/
             background-color: $color-gray-3 !important;
         }
 
+        &:focus {
+            outline: auto 2px Highlight !important;
```
michael-iden: I don't think I have ever seen the `Highlight` before. Is that custom for AIW?

jialin-he: No, it's what I was trying to simulate the "blue border" in firefox 😂

src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
eslawski: I was playing around with this a bit and I think I might have come up with something that prevents us from having to listen to the "tab" key. Instead I just leverage the `tabindex` so all thumbnails are tabbable to. I still needed to listen for the `tile-drawn` event, not sure there is a way around that. Let me know if this makes sense.

```        this.osdViewer.addOnceHandler("tile-drawn", () => {
            const referenceStripElements = document.querySelectorAll(".referencestrip .openseadragon-canvas");
            referenceStripElements.forEach((thumbnail, index) => {
                thumbnail.setAttribute("tabindex", "0");
                thumbnail.addEventListener("keydown", (event: KeyboardEvent) => {
                    if (event.keyCode === 13) {
                        this.osdViewer.goToPage(index)
                    }
                })
            })
        })```

File: src/app/shared/footer/footer.component.ts [hunk 4]
```diff
@@ -88,6 +89,27 @@ export class Footer {
...
+    }, false);
+  
+    document.addEventListener("msfullscreenchange", () => {
+      this.toggleFullscreen();
+    }, false);
+
+  }
+
+  private toggleFullscreen(): void {
+    this.isFullscreen = !this.isFullscreen
```
michael-iden: I wonder if there is any way we could centralize this logic between the nav and footer. In Vue it seems like we could use mixins but I don't know if we have the same construct available in Angular.

jialin-he: I'm thinking of putting it into a service and import the service in both components. I can try that after the meetings.

src/app/app.component.ts
eslawski: hmm so I just tested our jstor iiif viewer on firefox and noticed it suffers from the same problem. So it's likely a bigger issue. Interesting how openseadragon employs a similar fix. I am cool with going with these work arounds for now. Resolving this comment

src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
michael-iden: Does this end up working in firefox or do we still end up with the gray dotted border?

jialin-he: Yes, this ends up working in firefox.

src/app/asset-page/artstor-viewer/artstor-viewer.component.html
```diff
@@ -48,21 +59,14 @@
         <button class="btn btn--icon btn--prev" id="previousButton-{{ osdViewerId}}"><</button><span class="btn-divider"></span><button class="btn btn--icon btn--next" id="nextButton-{{ osdViewerId}}">></button>
         </div> -->
         <!-- Edge pagination arrow style -->
-        <div [class.aiw-hidden]="!isMultiView" class="asset-viewer__edge-arrows" id="imagePageButtons-{{ osdViewerId }}">
-        <button class="btn btn--icon btn--prev" id="previousButton-{{ osdViewerId}}"><i class="icon icon-prev--white"></i></button><button class="btn btn--icon btn--next" id="nextButton-{{ osdViewerId}}"><i class="icon icon-next--white"></i></button>
+        <div [class.aiw-hidden]="!isMultiView" class="asset-viewer__edge-arrows" id="imagePageButtons-{{ osdViewerId }}" tabindex="-1">
```
eslawski: I think there might be something going on with the previous button. I can't seem to get it to appear locally using this example: https://localhost:3000/#/asset/1002971919

src/app/asset-grid/thumbnail/thumbnail.component.ts
```diff
@@ -95,14 +98,16 @@ export class ThumbnailComponent implements OnInit, OnChanges {
...
     event.preventDefault()
     event.stopPropagation()
-
-    if (urlParams[0] === '/associated') {
-      this.angulartics.eventTrack.next({ properties: { event: 'view associated images', label: this.thumbnail.id } })
-    }
-    if (urlParams[0] === '/cluster') {
-      this.angulartics.eventTrack.next({ properties: { event: 'view cluster', label: this.thumbnail.id } })
+    // Don't navigate to other page in compare mode
+    if (!this.compareMode) {
```
eslawski: Have we talked with Nick about this behavior? Preventing this link from being clickable seems a little weird especially since you can tab to them. Wondering if we still wanted to show them in the first place?

src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
```diff
@@ -406,6 +406,29 @@ export class ArtstorViewerComponent implements OnInit, OnDestroy {
...
+                    let ZoomInElement: HTMLElement = <HTMLElement>document.getElementById(`imageButtons-${this.osdViewerId}`)
+                    ZoomInElement.focus()
+                }
+            })
+
+            // Loop through all canvas elements in the reference strip and add event listener to keydown enter for accessibility
+            for (let i = 2; i < canvasElements.length; i++) {
+                canvasElements.item(i).addEventListener("keydown", (event: KeyboardEvent) => {
+                    if (event.keyCode === 13) {
+                        this.osdViewer.goToPage(i-2)
```
michael-iden: There are a couple places where we have `2` hardcoded and it looks to be very intentional but I can't seem to figure out the intent. I'll keep reading through though.

eslawski: Pretty sure it is because the we are targeting the thumbnails in the reference strip. We essentially are counting the number of openseadragon canvases (since each thumbnail in the reference strip is technically a mini OSD viewer). When we query the canvases the first one works out being the main image being displayed, and the second one is the navigator component. Anything after those two should be thumbnails.

I wonder if we will get into trouble once we start adding more viewers via the compare mode. I think that will throw off the count we assume.

src/app/asset-page/artstor-viewer/artstor-viewer.component.html
```diff
@@ -21,20 +21,31 @@
...
-        <button class="btn btn--icon" [class.standalone-remove-icon]="state != 1" *ngIf="isFullscreen && assetCompareCount > 1" (click)="removeAsset.emit(asset)" aria-label="Remove from comparison" [ngbTooltip]="removeCompTooltip" placement="bottom"><i class="icon icon-remove-from-comparison"></i></button>
-      </div>
+        <div class="fullscreen-metadata" *ngIf="isFullscreen" [class.slideAway]="!showCaption">
+            <div class="vertical-center-wrap">
+                <i class="icon icon-direction icon-left" *ngIf="assetCompareCount == 1 && !quizMode && assetNumber > 1" title="Previous in results" tabindex="0" (click)="prevPage.emit()" (keydown.enter)="prevPage.emit()"></i>
+                <div class="title" [innerHtml]="asset.title"></div>
+                <i class="icon icon-direction icon-right" *ngIf="assetCompareCount == 1 && !quizMode && assetNumber < assetGroupCount" title="Next in results" tabindex="0" (click)="nextPage.emit()" (keydown.enter)="nextPage.emit()" (keydown.tab)="setFocusToCtrBtns()"></i>
```
eslawski: So if we are manually capturing the `keydown.tab` we might also need to implement the reverse in order to get the tab flow working properly when shift tabbing.

src/app/asset-page/artstor-viewer/artstor-viewer.component.html
michael-iden: Removing *ngIf="multiViewPage == multiViewCount" fixed the most immediate problem but that conditional may've been serving some other purpose that I didn't notice.

src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
michael-iden: Ahh 👍 now seeing the comment on line 411. I wonder if we could use something like https://openseadragon.github.io/docs/OpenSeadragon.World.html#getItemCount could help us out?

src/app/asset-page/artstor-viewer/artstor-viewer.component.html
jialin-he: The *ngIf is a leftover of previous implementation. I should remove it, doing that now.

src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
eslawski: Ah interesting idea to get the count from openseadragon. I think we also need to setup tabbing logic for the items in the reference strip though too. I threw out another potential solution here: https://github.com/ithaka/aiw-ui/pull/1207#discussion_r363758500 . Not sure we would be able to use `getItemCount` here though since we need to setup the `tabindex` on the thumbnails anyways.

src/app/asset-page/artstor-viewer/artstor-viewer.component.html
jialin-he: Oh yes, so more and more codes adding to it 😢

src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
jialin-he: One reason that I'm listening to the tab key is that we have to jump between "artstor-viewer" component and "asset-page" component. Normally (if I don't set up the listeners), we will be tabbing through all the interactive elements in one component before we can ever reach anything in the second.

src/app/asset-grid/thumbnail/thumbnail.component.ts
jialin-he: I haven't, I'm going to ask him when he's available.

src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
jialin-he: Oh my previous comment is under the assumption that we have all interactive elements with `tabIndex=0`, but if we set them to be `tabIndex=1`, `tabIndex=2`, etc, we might be able to accomplish some particular ordering. Worth trying.

PR #1209, opened 2020-01-06
SRCHX-237: Reported pages should have heading level 1.

rehanabbasi:
This PR addresses the accessibility issues reported in SRCHX-237.

### DIFF
--- a/src/sass/core/_typography.scss
+++ b/src/sass/core/_typography.scss
@@ -249,7 +249,7 @@ h1, h2, h3, h4, h5,
 
 h1, .h1 {
   @extend %Graphik-Regular-Web;
-  color: $typeAccent;
+  color: #444;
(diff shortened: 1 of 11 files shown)


src/sass/core/_typography.scss
```diff
@@ -249,7 +249,7 @@ h1, h2, h3, h4, h5,
 
 h1, .h1 {
   @extend %Graphik-Regular-Web;
-  color: $typeAccent;
+  color: #444;
```
momo-sa: Do we want to define this color as a variable to add a little context?

rehanabbasi: @momo-sa I agree, using sass variable would make more sense here. Change made in the recent commit `9af4e21`.

PR #1210, opened 2020-01-07
PLATO-225-focus-range-fix: Prevent invisible elements from getting focus in fullscreen mode

jialin-he:
Prevent elements that are invisible in fullscreen mode from getting focus :
- skip to main content link
- nav menu
- footer
- thumbnails in the drawer when the drawer is not open


src/app/_services/fullscreen.service.ts
```diff
@@ -0,0 +1,36 @@
+import { Injectable } from "@angular/core"
+
+@Injectable()
+export class FullScreenService {
```
eslawski: So I am not super familiar with Angular services, but I noticed a list of services we have listed here: https://github.com/ithaka/aiw-ui/blob/6002815ddf31785510b58039906c3a3e18a9ecd1/src/app/app.module.ts#L167

Does this new service need to be listed there?

src/app/app.module.ts
```diff
@@ -173,6 +174,7 @@ const APP_PROVIDERS = [
   AuthService,
   DatePipe,
   FlagService,
+  FullScreenService,
```
jialin-he: @eslawski I think I added the service here?

eslawski: ahh gotcha. Not sure how I missed that one!

src/app/app.component.ts
```diff
@@ -273,6 +276,24 @@ export class AppComponent {
...
+  }
+
+  private removeSkipToMainContent() {
+    let skipToMainContentButton = document.getElementById("skip-main-content-button")
+    skipToMainContentButton.tabIndex = -1
+  }
+
+  private addSkipToMainContent() {
+    let skipToMainContentButton = document.getElementById("skip-main-content-button")
+    skipToMainContentButton.tabIndex = 1
```
eslawski: I might be forgetting something, but was there a reason we are setting the `tabindex` instead of just showing/hiding the button using `ngIf` like we do with the nav/header?

jialin-he: Because the skip-to-main-content is written in html (not Angular component), where we can't use `*ngIf`.

jialin-he: Here is the code: https://github.com/ithaka/aiw-ui/pull/1210/files/b40cc3cd53daa508bdd5ce2e5bfbcc1f15fe4501#diff-5c26d2c8f8838f32ace87fdecd6344eaL33-R46

eslawski: I think that is an angular component right? It's just written in html form (not pug). I see a few angular specific directives in that block of code you referenced.

jialin-he: Ooooops, let me try to use `*ngIf` then!

jialin-he: Done!

src/app/app.component.ts
```diff
@@ -54,6 +54,8 @@ export class AppComponent {
   public test: any = {}
 
   public statusPageClient: any
+
+  private isFullscreen: boolean = false
```
eslawski: I think we can get rid of this guy right?

jialin-he: Yes and done!

eslawski (approved):
Looks good!

michael-iden (approved):
👍

PR #1211, opened 2020-01-07
PLATO-221-noticeable-outline: Create noticeable outline for interactive elements in viewer

jialin-he:
- Noticeable focus state for interactive elements in viewer
   - Blue border
   - Different background color
   - Different opacity for thumbnails in the reference strip
- Make pagination arrows, hide caption button, quiz mode button accessible

### DIFF
--- a/src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
+++ b/src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
@@ -69,6 +69,11 @@ $desktop: 992px;*/
             background-color: $color-gray-3 !important;
         }
 
+        &:focus {
+            outline: auto 2px Highlight !important;
+            outline: -webkit-focus-ring-color auto 5px !important;
+        }
+
         .icon {
             width: 24px !important;
             height: 24px !important;
@@ -144,6 +149,11 @@ $desktop: 992px;*/
 .asset-viewer__edge-arrows .btn:hover,
 .asset-viewer__edge-arrows .btn:active {
     background: rgba(66,66,66,1.0);
+}
+
+.asset-viewer__edge-arrows .btn:focus {
+    outline: auto 2px Highlight !important;
+    outline: -webkit-focus-ring-color auto 5px !important;
 }
 
 .asset-viewer__edge-arrows .btn--prev {
--- a/src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
+++ b/src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
@@ -69,11 +69,6 @@ $desktop: 992px;*/
             background-color: $color-gray-3 !important;
         }
 
-        &:focus {
-            outline: auto 2px Highlight !important;
-            outline: -webkit-focus-ring-color auto 5px !important;
-        }
-
         .icon {
             width: 24px !important;
             height: 24px !important;
@@ -149,11 +144,6 @@ $desktop: 992px;*/
 .asset-viewer__edge-arrows .btn:hover,
 .asset-viewer__edge-arrows .btn:active {
     background: rgba(66,66,66,1.0);
-}
-
-.asset-viewer__edge-arrows .btn:focus {
-    outline: auto 2px Highlight !important;
-    outline: -webkit-focus-ring-color auto 5px !important;
 }
 
 .asset-viewer__edge-arrows .btn--prev {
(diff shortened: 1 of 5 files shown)


src/app/asset-page/asset-page.component.pug
```diff
@@ -74,16 +74,16 @@
...
+          i.icon.icon-direction.icon-left([class.disabled]="!quizShuffle && assetNumber <= 1 ? true : false", title="Previous in results", tabindex="0", (tap)="showPrevAsset()", (keydown.enter)="showPrevAsset()")
+          i.icon.icon-direction.icon-right([class.disabled]="!quizShuffle && assetNumber >= totalAssetCount - restrictedAssetsCount ? true: false", title="Next in results", tabindex="0", (tap)="showNextAsset()", (keydown.enter)="showNextAsset()")
+        .shuffle-cntnr(*ngIf="quizMode", [ngClass]="[ quizShuffle ? 'active' : '' ]", tabindex="0", (tap)="toggleQuizShuffle()", (keydown.enter)="toggleQuizShuffle()", title="Toggle Asset Suffle")
           span.shuffle-btn
           | SHUFFLE {{ quizShuffle ? 'ON' : 'OFF' }}
-        .quiz-cntnr([ngClass]="[ quizMode ? 'active' : '' ]", (tap)="toggleQuizMode()", title="{{ quizMode ? 'Deactivate Quiz Mode' : 'Activate Quiz Mode' }}")
+        .quiz-cntnr([ngClass]="[ quizMode ? 'active' : '' ]", tabindex="0", (tap)="toggleQuizMode()", (keydown.enter)="toggleQuizMode()", (keydown.tab)="setFocusToCanvas()", title="{{ quizMode ? 'Deactivate Quiz Mode' : 'Activate Quiz Mode' }}")
```
eslawski: I don't think we have this `setFocusToCanvas` function yet, do we?

jialin-he: Oh we don't, going to remove it. Good eye!

src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
```diff
@@ -146,6 +151,11 @@ $desktop: 992px;*/
     background: rgba(66,66,66,1.0);
 }
 
+.asset-viewer__edge-arrows .btn:focus {
+    outline: auto 2px Highlight !important;
```
eslawski: It's a little curious that we are manually trying to reproduce the blue boarding chrome provides. Is there a reason why we don't get this for free?

gbassham: Do other browsers implement the same border? I thought Firefox was different, could be wrong.

michael-iden: Browsers do have different focus outlines but I think it would be best if we could try to keep things to be the browser default? But Evan and I were talking and noticed that open seadragon might override the default browser CSS and trying to re-override it with the default might be tricky

jialin-he: Firefox's default border is dotted lines. The dotted line cannot be viewed in fullscreen mode, that's why I manually add blue border to it.

jialin-he: For chrome, elements created by openseadragon and elements in the `artstor.viewer.component.html` don't get the blue border for free.

eslawski: hmm interesting. I am wondering if the default bordering is even what we are after from a UX perspective. I know on Chrome we get the blue bordering, which we are trying to replicate, but on firefox we get the dotted outline, which might not be great for accessibility. Maybe we need to talk with Nick to see if there is anything better we can do to create uniformity instead of trying to rehardcoded defaults.

Potentially unrelated, but I found some other places where we are explicitly turning off some outlines. 
https://github.com/ithaka/aiw-ui/blob/461b7b24edd955479bd240438f39de062a735ec5/src/sass/modules/_viewer.scss#L2-L6

eslawski: I noticed that for this particular css rule we are actually dealing iwth the pagination button though. I noticed that the css that is removing that border is probably related to this: https://github.com/ithaka/aiw-ui/blob/461b7b24edd955479bd240438f39de062a735ec5/src/assets/styleguide/styles/all.css#L1639

At least that is what I can see in the DOM

jialin-he: @eslawski That's for non-focus state. But even we remove that line, the default border is still missing. I tried to use default borders before I turn to the solution of adding hardcoded borders, but it didn't work. I can dig more though.

eslawski (approved):

michael-iden (approved):

PR #1213, opened 2020-01-08
PLATO-487: Make thumbnails in the reference strip accessible with keyboard

jialin-he:
Add event listeners to thumbnails in the reference strip so that hitting enter on the thumbnail will paginate to that image.

Note: there is a hardcoded `i-2` because the first element with `openseadragon-canvas` class will always be the main canvas, the second element will be the navigator (this is true even in fullscreen mode where we don't show the navigator). The thumbnails start with the third element in the collection. And it is true in compare mode, too.

### DIFF
--- a/src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
+++ b/src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
@@ -406,6 +406,20 @@ export class ArtstorViewerComponent implements OnInit, OnDestroy {
             this.osdViewer.destroy();
         });
 
+        this.osdViewer.addOnceHandler("tile-drawn", () => {
+            // Get all canvas elements, first one is the main canvas, second one is the navigator
+            // The canvas in the reference strip starts as the third one in the collection
+            let canvasElements = document.getElementsByClassName("openseadragon-canvas")
+            // Loop through all canvas elements in the reference strip and add event listener to keydown enter for accessibility
+            for (let i = 2; i < canvasElements.length; i++) {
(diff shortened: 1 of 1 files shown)


src/app/asset-page/artstor-viewer/artstor-viewer.component.ts
```diff
@@ -406,6 +406,20 @@ export class ArtstorViewerComponent implements OnInit, OnDestroy {
             this.osdViewer.destroy();
         });
 
+        this.osdViewer.addOnceHandler("tile-drawn", () => {
+            // Get all canvas elements, first one is the main canvas, second one is the navigator
+            // The canvas in the reference strip starts as the third one in the collection
+            let canvasElements = document.getElementsByClassName("openseadragon-canvas")
```
eslawski: Wondering if we can use a more specific selector to pick out only the `openseadragon-canvas`s in the `referencestrip`? That way we wouldn't need always start at i = 2

jialin-he: Could do something similar to this https://github.com/ithaka/aiw-ui/pull/1207/files#diff-1373abb83f1b3bd7243873880bf670f8R12 ?

jialin-he: I use a more specific selector now to get rid of the hardcoded `2`

michael-iden (approved):

eslawski (approved):

eslawski:
Looks good! FYI I think this might still be based on `PLATO-221-noticeable-outline`

PR #1216, opened 2020-01-09
PLATO-488: Create correct tab order for Artstor viewer

jialin-he:
- In fullscreen mode, the tab order is:
   - Main canvas
   - Zoom in, zoom out, fit to view
   - Compare, exit
   - Page arrows
   - Thumbnails in the reference strip if it is multi-view
   - Pagination buttons
   - Hide caption, quiz mode toggle button
- Make sure the newly added `tabindex` doesn't interfere the original tab order in non-fullscreen mode

### DIFF
--- a/src/app/asset-page/artstor-viewer/artstor-viewer.component.html
+++ b/src/app/asset-page/artstor-viewer/artstor-viewer.component.html
@@ -29,10 +29,10 @@
     <!--Viewer Buttons-->
     <div *ngIf="!thumbnailMode">
       <div class="button-group asset-viewer__buttons" id="imageButtons-{{ osdViewerId }}">
-        <button class="btn btn--icon btn--zoomIn" [class.aiw-hidden]="state != 1" id="zoomIn-{{ osdViewerId}}" [ngbTooltip]="zoomInTooltip" aria-label="Zoom in" placement="bottom"><i class="icon icon-zoom-in"></i></button>
-        <button class="btn btn--icon btn--zoomOut" [class.aiw-hidden]="state != 1" id="zoomOut-{{ osdViewerId }}" [ngbTooltip]="zoomOutTooltip" aria-label="Zoom out" placement="bottom"><i class="icon icon-zoom-out"></i></button>
-        <button class="btn btn--icon btn--zoomFit" [class.aiw-hidden]="state != 1" id="zoomFit-{{ osdViewerId }}" [ngbTooltip]="fitToViewTooltip" aria-label="Fit to view" placement="bottom"><i class="icon icon-fit-to-view"></i></button>
-        <button class="btn btn--icon btn--fullScreen" [class.standalone-present-icon]="state != 1" *ngIf="!isFullscreen" id="fullScreen-{{ osdViewerId }}" aria-label="Full screen" (click)="togglePresentationMode()" [ngbTooltip]="presentTooltip" placement="bottom"><i class="icon icon-present"></i></button>
+        <button class="btn btn--icon btn--zoomIn" [class.aiw-hidden]="state != 1" id="zoomIn-{{ osdViewerId}}" [ngbTooltip]="zoomInTooltip" aria-label="Zoom in" placement="bottom" tabindex="1"><i class="icon icon-zoom-in"></i></button>
+        <button class="btn btn--icon btn--zoomOut" [class.aiw-hidden]="state != 1" id="zoomOut-{{ osdViewerId }}" [ngbTooltip]="zoomOutTooltip" aria-label="Zoom out" placement="bottom" tabindex="1"><i class="icon icon-zoom-out"></i></button>
+        <button class="btn btn--icon btn--zoomFit" [class.aiw-hidden]="state != 1" id="zoomFit-{{ osdViewerId }}" [ngbTooltip]="fitToViewTooltip" aria-label="Fit to view" placement="bottom" tabindex="1"><i class="icon icon-fit-to-view"></i></button>
+        <button class="btn btn--icon btn--fullScreen" [class.standalone-present-icon]="state != 1" *ngIf="!isFullscreen" id="fullScreen-{{ osdViewerId }}" aria-label="Full screen" (click)="togglePresentationMode()" [ngbTooltip]="presentTooltip" placement="bottom" tabindex="1"><i class="icon icon-present"></i></button>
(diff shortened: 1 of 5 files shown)


src/app/asset-page/asset-page.component.pug
```diff
@@ -74,16 +74,16 @@
 
       .col-md-12.quiz-caption-cntnr(*ngIf="isFullscreen", [ngClass]="[ showAssetCaption ? '' : 'cntnr-bg' ]")
         ang-promo-tooltip(*ngIf="user.isLoggedIn && !quizModeTTDismissed && studyMode", [options]="quizModeTooltipOpts", (close)="closeQuizModeTooltip()")
-        .caption-btn-cntnr([ngClass]="[ showAssetCaption ? 'active' : '' ]", tabindex="0", (click)="showAssetCaption = !showAssetCaption", (keydown.enter)="showAssetCaption = !showAssetCaption", title="{{ showAssetCaption ? 'Hide Asset Meta Information' : 'Show Asset Meta Information' }}")
+        .caption-btn-cntnr([ngClass]="[ showAssetCaption ? 'active' : '' ]", tabindex="6", (click)="showAssetCaption = !showAssetCaption", (keydown.enter)="showAssetCaption = !showAssetCaption", title="{{ showAssetCaption ? 'Hide Asset Meta Information' : 'Show Asset Meta Information' }}")
```
michael-iden: Forgive me if this has already been discussed on other pull requests but I wonder if we could fix the tab ordering without using the tabindex. It seems that it might be rather tricky to maintain. Moving around the ordering of items in the dom without any tab indexing would probably be the most clear to me but it may not be possible/feasible.

jialin-he: Oh yes, that would be perfect if we can do that. But the problem is, we have two components: `artstor-viewer` and `asset-page`. `artstor-viewer` is a child component of `asset-page`. The viewer, thumbnails in the reference strip, zoom buttons, and pagination buttons are in `arstor-viewer`, while the compare and exit button, and quiz buttons are in `asset-page` component for implementation reasons. If I give `tabindex=0` to all of them, the elements inside `arstor-viewer` will always get focus first (not the order we want).

src/app/asset-page/asset-page.component.pug
```diff
@@ -252,47 +252,48 @@
...
                 .form-group-feedback {{ 'ASSET_PAGE.EDIT_DETAILS_FORM.SAVE_ERROR' | translate }}
               .btn-row
-                button.btn.btn-secondary((click)="showDeletePCModal = true", type="button", tabindex="6", [class.loading]="deleteLoading") {{ 'ASSET_PAGE.EDIT_DETAILS_FORM.DELETE_BTN' | translate }}
-                button.btn.btn-primary(type="submit", [class.loading]="isProcessing", tabindex="6") {{ 'ASSET_PAGE.EDIT_DETAILS_FORM.SAVE_BTN' | translate }}
+                button.btn.btn-secondary((click)="showDeletePCModal = true", type="button", [class.loading]="deleteLoading") {{ 'ASSET_PAGE.EDIT_DETAILS_FORM.DELETE_BTN' | translate }}
+                button.btn.btn-primary(type="submit", [class.loading]="isProcessing") {{ 'ASSET_PAGE.EDIT_DETAILS_FORM.SAVE_BTN' | translate }}
 
     .fullscreen-btns.button-group(*ngIf="isFullscreen", [class.fade-out]="showAssetDrawer")
-      button.btn.btn-light(*ngIf="!quizMode", (click)="toggleAssetDrawer(!showAssetDrawer)")
+      button.btn.btn-light(*ngIf="!quizMode", tabindex="7", (click)="toggleAssetDrawer(!showAssetDrawer)")
```
eslawski: The PR description says the compare and exit full screen buttons are supposed to come after the zoom in/out and fit to view buttons (when in full screen). I am noticing them coming after the quiz mode.

src/app/asset-page/artstor-viewer/artstor-viewer.component.html
```diff
@@ -49,15 +49,15 @@
...
         <b>{{ multiViewPage }}</b> of <b>{{ multiViewCount }}</b>
         <button *ngIf="hasMultiViewHelp()" (click)="multiViewHelp.emit()" class="help-icon">?</button>
         </div>
         <div class="fullscreen-metadata" *ngIf="isFullscreen" [class.slideAway]="!showCaption">
-        <div class="vertical-center-wrap"><i class="icon icon-direction icon-left" *ngIf="assetCompareCount == 1 && !quizMode" [class.disabled]="assetNumber <= 1" title="Previous in results" tabindex="0" (click)="prevPage.emit()" (keydown.enter)="prevPage.emit()"></i>
-            <div class="title" [innerHtml]="asset.title"></div><i class="icon icon-direction icon-right" *ngIf="assetCompareCount == 1 && !quizMode" [class.disabled]="assetNumber >= assetGroupCount" title="Next in results" tabindex="0" (click)="nextPage.emit()" (keydown.enter)="nextPage.emit()"></i>
+        <div class="vertical-center-wrap"><i class="icon icon-direction icon-left" *ngIf="assetCompareCount == 1 && !quizMode" [class.disabled]="assetNumber <= 1" title="Previous in results" tabindex="5" (click)="prevPage.emit()" (keydown.enter)="prevPage.emit()"></i>
```
eslawski: I appears I can tab to these arrows if they are disabled. Did we want to hide them?

src/app/asset-page/asset-page.component.pug
jialin-he: Ah good catch! I gave it a `tabindex=2` first, and then I wanted to change the old elements with `tabindex=2` to `tabidex=7` and I mistakenly changed them, too. 😂

src/app/asset-page/artstor-viewer/artstor-viewer.component.html
jialin-he: Yes, it's in another branch. PLATO-486 it is

eslawski:
I think the compare workflow might need some work. I am noticing that I can only tab to the `Artstor` badges. We might need to discuss with Nick, but I would think we want to be able to highlight the thumbnails in the compare mode strip, and then select them with ENTER to add them to the compare mode.

We also might want to set direct focus on the compare strip when the option is selected.

In addition, once you select an image and are comparing more than one image, the accessibility seems to break. Seems like you should be able to tab to each of the images and their respective toolbars. 

Either way it seems like there is a fair amount of work, and wondering if the use of `tabindex` starts to break down here...

PR #1217, opened 2020-01-10
PLATO-488: Fix tab ordering of the viewer

jialin-he:
Now the tab order is:
 - zoom buttons
 - multi-view pagination button (multi-views only)
 - main canvas
 - thumbnails in the reference strip (multi-views only)
 - results pagination button
 - hide caption & quiz mode button
 - compare & exit button
 - interactive elements in the drawer (if drawer opened)

### DIFF
--- a/src/app/asset-page/artstor-viewer/artstor-viewer.component.html
+++ b/src/app/asset-page/artstor-viewer/artstor-viewer.component.html
@@ -1,4 +1,45 @@
 <div class="asset-viewer" [class.fullscreen]="isFullscreen" [class.multi-view]="isMultiView" (contextmenu)="disableContextMenu($event)">
+  <!--Viewer Buttons-->
+  <div *ngIf="!thumbnailMode">
+    <div class="button-group asset-viewer__buttons" id="imageButtons-{{ osdViewerId }}">
+      <button class="btn btn--icon btn--zoomIn" [class.aiw-hidden]="state != 1" id="zoomIn-{{ osdViewerId}}" [ngbTooltip]="zoomInTooltip" aria-label="Zoom in" placement="bottom"><i class="icon icon-zoom-in"></i></button>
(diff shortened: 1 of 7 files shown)


src/app/asset-page/artstor-viewer/artstor-viewer.component.html
```diff
@@ -1,5 +1,37 @@
 <div class="asset-viewer" [class.fullscreen]="isFullscreen" [class.multi-view]="isMultiView" (contextmenu)="disableContextMenu($event)">
-    <!-- Large thumbnail fallback
+  <!--Viewer Buttons-->
```
eslawski: This block of html was practically moved as-is from the bottom, up here to the top. This ensures the viewer buttons (which are overlaid on top of the viewer) come BEFORE the viewer in terms of tab order.

There is one exception however and that is the `fullscreen-metadata` div (still at the bottom). We had to extract that div and leave it where it was, because we want that tab order to come AFTER the viewer. That html element still remains as-is too.

src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
```diff
@@ -33,23 +33,14 @@ $desktop: 992px;*/
     overflow: hidden;
 }
 
-.asset-viewer:hover .asset-viewer__buttons  {
```
eslawski: If you were not hovering over the viewer the zoom/in out controls etc would auto hide themselves. This is not very accessibility friendly, and it was hiding the focus. We talked with Nick and determined that we will just always show the controls from here on out.

src/app/asset-page/artstor-viewer/artstor-viewer.component.scss
```diff
@@ -116,24 +107,20 @@ $desktop: 992px;*/
...
-    opacity: 1;
-}
-
 .asset-viewer__edge-arrows .btn {
     pointer-events: all;
     width: 42px;
     height: 42px;
     border-radius: 50%;
     background: rgba(66,66,66,0.6);
+    z-index: 9999;
```
eslawski: For whatever reason moving the viewer buttons before the iiif viewer causes the pagination buttons to be covered up. Adding a z-index here ensures they are overlaid on top of the viewer

eslawski (approved):
Not sure if mine counts since I worked on some of it, but this is looking good!

michael-iden (approved):

PR #1218, opened 2020-01-13
PALS-170: Document title element missing/empty

crystalfinch:
Resolves [PALS-170](https://jira.jstor.org/browse/PALS-170)

Title is now being set on Home, Register and Log in pages

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -5,7 +5,7 @@ import { Component, ViewEncapsulation, PLATFORM_ID, Inject, Renderer2 } from '@a
 import { DOCUMENT } from '@angular/common'
 import { Angulartics2GoogleTagManager } from 'angulartics2/gtm'
 import { Title, Meta } from '@angular/platform-browser'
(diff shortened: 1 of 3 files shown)


src/app/app.component.ts
```diff
@@ -243,6 +239,15 @@ export class AppComponent {
     })).subscribe()
   }
 
+  updateTitle(event: RouterEvent) {
+      let event_url_array = event.url.split('/')
+      if (event_url_array && (event_url_array.length > 1)){
+        if (event_url_array[1] === 'home' || event_url_array[1] !== 'asset') {
+          this.titleService.setTitle(this.title)
+        }
+      }
```
momo-sa: Isn't that equivalent to:

```javascript
      if (event_url_array && event_url_array.length > 1 &&
          (event_url_array[1] === 'home' || event_url_array[1] !== 'asset') {
        this.titleService.setTitle(this.title)
      }
```

I'm not sure it's worth making that change. Depends on what you find more readable.

crystalfinch: Yep, it was for readability!

PR #1221, opened 2020-01-14
PLATO-227: Fix screenreader message error on group presentation mode

michael-iden:
Resolves [PLATO-227](https://jira.jstor.org/browse/PLATO-227)

Fixing error generated when asset may not have been loaded by the time we want to use it for the screenreader message.

### DIFF
--- a/src/app/shared/group-title/group-title.component.ts
+++ b/src/app/shared/group-title/group-title.component.ts
@@ -88,10 +88,12 @@ import { ImageGroup } from 'datatypes'
         queryParams['w'] = this.ig.items[0].zoom.pointWidth
         queryParams['h'] = this.ig.items[0].zoom.pointHeight
       }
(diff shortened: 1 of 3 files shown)


src/app/shared/group-title/group-title.component.ts
```diff
@@ -88,10 +88,12 @@ import { ImageGroup } from 'datatypes'
         queryParams['w'] = this.ig.items[0].zoom.pointWidth
         queryParams['h'] = this.ig.items[0].zoom.pointHeight
       }
-      // Enter fullscreen - must fire within click binding (Firefox, Safari)
-      this._toolbox.requestFullScreen()
       // Route to viewer
-      this._router.navigate(['/asset', id, queryParams]);
+      this._router.navigate(['/asset', id, queryParams])
+        .then(() => {
```
jialin-he: Ahhhh good to know we can do that! This should solve lots of potential problems!

PR #1231, opened 2020-01-22
SRCHX-242: Add visually hidden label on Search input

willschoener:


src/app/shared/search/search.component.pug
```diff
@@ -1,6 +1,7 @@
 .search-wrap
   .form-group.row
     .input-group([ngClass]="{'col-md-7': allowSearchInRes}")
+      label.visually-hidden(for="inputSearchTerm") Search Artstor
```
rehanabbasi: We believe we already have a class `.sr-only` defined for such use cases.

PR #1232, opened 2020-01-23
SRCHX-429: Show outline on showdow for pagination controls

rehanabbasi:
The PR ensures that we show the outline for the pagination control once it gains focus.

### DIFF
--- a/src/app/shared/pagination/pagination.component.scss
+++ b/src/app/shared/pagination/pagination.component.scss
@@ -80,8 +80,9 @@ button.page-link {
     }
 
     &:active, &:focus {
-        outline: none;
         box-shadow: none;
+        outline: auto;
+        outline-color: -webkit-focus-ring-color;
     }
 }
 


src/app/shared/pagination/pagination.component.scss
```diff
@@ -80,8 +80,9 @@ button.page-link {
     }
 
     &:active, &:focus {
-        outline: none;
         box-shadow: none;
+        outline: auto;
+        outline-color: -webkit-focus-ring-color;
```
michael-iden: We found that focus ring color sometimes didn't work for firefox when the -webkit-focus-ring-color variable was used. We have used `Highlight` which is a system color and that seems to work but there is some discrepancy about if it should be used or not. Here is the stackoverflow we found to get "Highlight", but then marks the answer as "Do not use" https://stackoverflow.com/questions/7538771/what-is-webkit-focus-ring-color

But going directly to the documentation says "Authors may also use these keywords at any time, but should be careful to use the colors in matching background-foreground pairs to ensure appropriate contrast..." https://www.w3.org/TR/css-color-4/#css-system-colors

🤷‍♂

PR #1234, opened 2020-01-23
PALS-115: Register Page: Error Messages Aren't Announced via Screen Reader

crystalfinch:
Apply focus to error messages on Register page after a timeout so VoiceOver doesn't announce them too soon.

### DIFF
--- a/src/app/register/register.component.ts
+++ b/src/app/register/register.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit } from '@angular/core'
+import { Component, ElementRef, OnInit, ViewChild } from '@angular/core'
 import { Router, ActivatedRoute } from '@angular/router'
 import { formGroupNameProvider } from '@angular/forms/src/directives/reactive_directives/form_group_name'
 import { FormGroup, FormControl, FormBuilder, Validators } from '@angular/forms'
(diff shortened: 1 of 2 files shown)


src/app/register/register.component.ts
```diff
@@ -100,9 +104,35 @@ export class RegisterComponent implements OnInit {
...
+  private focusError(error: boolean, el: ElementRef) {
+    if (error) {
+      if (el && el.nativeElement) {
+        this.focusErrorElement(el.nativeElement)
+      }
+    }
+  }
+
+  private focusErrorElement(el: HTMLElement) {
+    setTimeout(() => {
```
momo-sa: Is it feasible to use retries instead of timeouts to ensure the page is ready as quickly as possible? Adding an extra second seems like a last resort.

crystalfinch: @momo-sa The use of the settimeout here is not to ensure the page is ready. There is code in other components that set focus after a timeout delay of 100ms. When testing with VoiceOver, the delay of 1000ms was the sweet spot that made it so the narrator didn't stumble on words by being re-focused too quickly. Hope that makes sense! I can do a demo for you if you like.

PR #1239, opened 2020-01-29
PLATO-496: Automated accessibility fixes for image group page

michael-iden:
Resolves [PLATO-496](https://jira.jstor.org/browse/PLATO-496)

Fixes automated accessibility errors generated from the QA gem.

### DIFF
--- a/src/app/asset-grid/asset-grid.component.pug
+++ b/src/app/asset-grid/asset-grid.component.pug
@@ -123,9 +123,9 @@
                 [largeThmbView]="largeThmbView"
               )
             .mobile--hide
-              input.check(tabIndex="3", type="checkbox", [checked]="isSelectedAsset(asset) > -1", [ngStyle]="{ 'opacity': isSelectedAsset(asset) > -1 ? '1' : '0' }", (change)="editMode = true; selectAsset(asset);")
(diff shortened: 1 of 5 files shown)


src/app/asset-grid/asset-grid.component.pug
```diff
@@ -123,9 +123,9 @@
                 [largeThmbView]="largeThmbView"
               )
             .mobile--hide
-              input.check(tabIndex="3", type="checkbox", [checked]="isSelectedAsset(asset) > -1", [ngStyle]="{ 'opacity': isSelectedAsset(asset) > -1 ? '1' : '0' }", (change)="editMode = true; selectAsset(asset);")
+              input.check([attr.aria-label]="'Select asset \"' + asset.name + '\"'", tabIndex="3", type="checkbox", [checked]="isSelectedAsset(asset) > -1", [ngStyle]="{ 'opacity': isSelectedAsset(asset) > -1 ? '1' : '0' }", (change)="editMode = true; selectAsset(asset);")
             .mobile--show
-              input.check-mobile(type="checkbox", [checked]="isSelectedAsset(asset) > -1", (change)="editMode = true; selectAsset(asset);")
```
michael-iden: This is real ugly. Please lmk if there is a better way to do the label construction.

eslawski: Would it be overkill to extract the string building into a function/variable on `asset-grid.component`? Might be able to use ES6 string templates there too?

gbassham: I think Evan is right, I would put this into a function. Easier to maintain if it has to be changed in the future.

michael-iden: Done!

PR #1245, opened 2020-02-10
PLATO-466: Confirm and build page of commands for VISIBLE Image Viewer Keyboard Instructions

jialin-he:
Resolves PLATO-466

## Description

- Build a support page that lists image viewer commands: https://openseadragon.github.io/examples/ui-keyboard-navigation/
- Same style as workspace notes keyboard navigation: https://www.jstor.org/support/text-editor-help/

## Checks

- [ ] Yes
- [ ] Not yet
- [x] Not applicable

- [ ] Yes
- [x] Not yet
- [ ] Not applicable

**Browsers tested on:**

- [x] Chrome
  - [ ] Mobile
- [ ] Safari
  - [ ] Mobile
- [ ] Firefox
  - [ ] Mobile
- [ ] Edge
- [ ] IE11
- [ ] Not applicable

### DIFF
--- a/src/app/support-page/support-page.component.pug
+++ b/src/app/support-page/support-page.component.pug
@@ -8,31 +8,31 @@
     tbody
       tr
         td Move viewport up
-        td W, Up Arrow
+        td w, Up Arrow
       tr
         td Move viewport down
-        td S, Down Arrow
+        td s, Down Arrow
       tr
         td Move viewport left
-        td A, Left Arrow
+        td a, Left Arrow
       tr
         td Move viewport right
-        td D, Right arrow
+        td d, Right arrow
       tr
         td Zoom / Move viewport home
         td 0
       tr
         td Zoom viewport out
-        td - / _, Shift + S, Shift + Down arrow
+        td - / _, Shift + s, Shift + Down arrow
       tr
         td Zoom viewport in
-        td = / +, Shift + W, Shift + Up arrow
+        td = / +, Shift + w, Shift + Up arrow
       tr
         td Rotate clockwise
-        td R
+        td r
       tr
         td Rotate counterclockwise
-        td Shift + R
+        td Shift + r
       tr
         td Flip horizontally
-        td F
+        td f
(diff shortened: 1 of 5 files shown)


src/app/support-page/support-page.component.pug
```diff
@@ -0,0 +1,38 @@
...
+        td 0
+      tr
+        td Zoom viewport out
+        td - / _, Shift + S, Shift + Down arrow
+      tr
+        td Zoom viewport in
+        td = / +, Shift + W, Shift + Up arrow
+      tr
+        td Rotate clockwise
+        td R
```
michael-iden: We may want to use use a lower case character here and with the other commands where users do not have have to  hit the shift key to perform the action. 

```suggestion
        td r
```

src/app/support-page/support-page.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'app-support-page',
+  templateUrl: './support-page.component.pug',
+  styleUrls: ['./support-page.component.scss']
+})
+export class SupportPageComponent implements OnInit {
```
michael-iden: We may be able to get rid of the constructor and OnInit if we don't implement OnInit 

```suggestion
export class SupportPageComponent {
```

src/app/support-page/support-page.component.pug
jialin-he: Hahaha I use upper case because I see the letters are upper case in the example: https://www.jstor.org/support/text-editor-help/

michael-iden: We may want to change the directions to be lowercase there too, but I suggested it here because `w` and `W` perform different actions. This may be a question for Nick though

jialin-he: @sandsng can you provide some suggestions?

sandsng: I like the suggestion of switching to lowercase. Makes sense! Thanks!

michael-iden (approved):
👍

NazimHAli (approved):

PR #1250, opened 2020-02-20
PALS-276: Add Brand survey on Artstor

satya-achanta-venkata:
Resolves PALS-276

## Description

Add brand survey on Artstor using ethn.io

## Checks

- [ ] Yes
- [ ] Not yet
- [x] Not applicable

- [x] Yes
- [ ] Not yet
- [ ] Not applicable

**Browsers tested on:**

- [x] Chrome
  - [ ] Mobile
- [x] Safari
  - [ ] Mobile
- [x] Firefox
  - [ ] Mobile
- [ ] Edge
- [ ] IE11
- [ ] Not applicable

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -190,9 +190,6 @@
   <!-- For cookie banner -->
   <script src="https://static.ithaka.org/gdpr-banner/js/banner.js" type="text/javascript"></script>
 
-  <!-- Ethnio Activation Code -->
-  <script type="text/javascript" language="javascript" src="//ethn.io/49123.js" async="true" charset="utf-8"></script>
-
   <p id="toast-live-region" class="sr-only" role="region" aria-live="polite" aria-atomic="true"></p>
 
 </body>
--- a/src/index.html
+++ b/src/index.html
@@ -190,6 +190,9 @@
   <!-- For cookie banner -->
   <script src="https://static.ithaka.org/gdpr-banner/js/banner.js" type="text/javascript"></script>
 
+  <!-- Ethnio Activation Code -->
+  <script type="text/javascript" language="javascript" src="//ethn.io/49123.js" async="true" charset="utf-8"></script>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -190,6 +190,9 @@
   <!-- For cookie banner -->
   <script src="https://static.ithaka.org/gdpr-banner/js/banner.js" type="text/javascript"></script>
 
+  <!-- Ethnio Activation Code -->
```
eslawski: So we I know we have added/removed the Ethino survey a couple times on artstor. This is an example how we have done it in the past: https://github.com/ithaka/aiw-ui/pull/1162/files . I think it has to do with the fact that artstor is server side rendered.

Wondering if we should be using the `scriptStore`?

eslawski:
Didn't we also want to exclude it when the denial modal was showing?

src/index.html
satya-achanta-venkata: The actual requirement is to display Ethnio survey in all the pages, meaning once per session but wherever the user lands first, we should be displaying. Since index.html is the main html file that is being used irrespective of which view is displayed ( correct me if I am wrong ), I added the script in index.html. I looked at the PR you included above, which requires adding ScriptService in all the components individually. Personally I prefer index.html because it comes into scene every time when any view displayed.

satya-achanta-venkata:
@eslawski  Spoke to Mary about this, and she mentioned that there were no denial pages in the Artstor ( not Modals but pages ), so we decided to display in all the pages.

src/index.html
eslawski: I think `app.component.ts` is the top level angular component. Wondering if we could load ethnio from the script loader in there instead.

satya-achanta-venkata: Yes, I tried to find whether I can use the script in app component. Since, app component doesn't have a view, I found where exactly the selector 'app-root' is being used in the project. Its inside <body> tag of index.html, we are referencing app-root selector which is tied to app component.

PR #1257, opened 2020-03-03
PALS-294: AIW: Temporarily remove login-to-download requirement

crystalfinch:
PALS-294

## Description

If a user has access via an ADL Institution, let them download an image (after accepting T&Cs) without logging into an individual account.

## Checks

- [ ] Yes
- [ ] Not yet
- [x] Not applicable

- [ ] Yes
- [ ] Not yet
- [x] Not applicable

**Browsers tested on:**

- [x] Chrome
  - [ ] Mobile
- [ ] Safari
  - [ ] Mobile
- [ ] Firefox
  - [ ] Mobile
- [ ] Edge
- [ ] IE11
- [ ] Not applicable

### DIFF
--- a/src/app/_services/flag.service.ts
+++ b/src/app/_services/flag.service.ts
@@ -91,7 +91,8 @@ export interface FeatureFlags {
   solrMetadata?: boolean,
   // detailViews?: boolean,
   exportGoogle?: boolean,
-  relatedResFlag?: boolean
+  relatedResFlag?: boolean,
+  requireLoginForAdlDownload?: boolean
 }
 
 interface FlagServiceResponse {
--- a/src/app/asset-page/asset-page.component.pug
+++ b/src/app/asset-page/asset-page.component.pug
@@ -89,7 +89,6 @@
     //- Asset Metadata Column
     .col-md-4(*ngIf="assets[0]",[class.hidden]="isFullscreen")
       .py-3
-        //- If the user has not accepted the agreement, show modal when clicked, otherwise download immediately
         .btn-row
           //- Edit
           //- * (mouseup) is a Firefox fix for firing the copy event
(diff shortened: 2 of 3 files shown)


src/app/asset-page/asset-page.component.pug
```diff
@@ -118,43 +117,19 @@
...
+            ng-container(*ngIf="!user.isLoggedIn && !assets[0].publicDownload")
+              ng-container(*ngTemplateOutlet="falseDoorDownload")
+          ng-container(*ngIf="!requireLoginForAdlDownload")
+            ng-container(*ngIf="assets[0].publicDownload")
+              ng-container(*ngTemplateOutlet="downloadDropdown")
+            ng-container(*ngIf="!assets[0].publicDownload")
+              ng-container(*ngIf="(user?.status || user.isLoggedIn)")
+                ng-container(*ngTemplateOutlet="downloadDropdown")
+              ng-container(*ngIf="!(user?.status || user.isLoggedIn)")
+                ng-container(*ngTemplateOutlet="falseDoorDownload")
```
eslawski: There are quite a few branches to the `if` condition here. Wondering if we could consolidate and simplify this a bit. It seems like the template really has two things it cares about 1) should I show the `downloadDropdown` or 2) the `falseDoorDownload` button? Maybe we can move some of the boolean logic to a function that can tell us `canUserDownload` (or something like that) and use it here in the template? That can take into consideration things like the `publicDownload`, `downloadLink`, `requireLoginForAdlDownload`, `isLoggedIn`, etc. etc.

I also feel like there is an opportunity to consolidate some of the boolean logic, but I am not exactly sure how that would look right now. I help take a look too if you want!

src/app/_services/flag.service.ts
```diff
@@ -91,7 +91,8 @@ export interface FeatureFlags {
   solrMetadata?: boolean,
   // detailViews?: boolean,
   exportGoogle?: boolean,
-  relatedResFlag?: boolean
+  relatedResFlag?: boolean,
+  requireLoginForAdlDownload?: boolean
```
eslawski: So has this new flag already been initialized to `false` some how?

crystalfinch: Yes, it's set to false in the test-deployed aiw-ui.json file here: https://github.com/ithaka/artstor-flags/pull/145/files

src/app/asset-page/asset-page.component.pug
crystalfinch: I pushed a much better solution in b5f3fcd 👍

src/app/asset-page/asset-page.component.ts
```diff
@@ -453,6 +455,27 @@ export class AssetPage implements OnInit, OnDestroy {
     this._group.hasPrivateGroups()
   } // OnInit
 
+  get showDownloadDropdown() {
+    const isLoggedIn = this.user && this.user.isLoggedIn;
+    const publicDownload = this.assets && this.assets[0] && this.assets[0].publicDownload;
+    const downloadLink = this.assets && this.assets[0] && this.assets[0].downloadLink;
+    const userStatus = this.user && this.user.status;
```
eslawski: The `userStatus` here is really an indicator if the user is "institutionally authed" right? Wondering if renaming it to reflect such would help the reading of the if conditions we have below.

src/app/asset-page/asset-page.component.ts
```diff
@@ -453,6 +455,27 @@ export class AssetPage implements OnInit, OnDestroy {
...
+      if (!publicDownload) {
+        return userStatus || isLoggedIn;
+      }
+
+      return true;
+    }
+
+    return !!(isLoggedIn || publicDownload);
+  }
+
```
eslawski: This already reads a lot better than when it lives in the template! Wondering if we can reduce it even further. If I am following things correctly I think that `publicDownload` always seems to resolve be `true` (assuming we actually have the `downloadLink`. Wondering if we can move that upwards to simplify things.

Does this seem equivalent?
```if(!downloadLink) {
	return false;
}

if(publicDownload) {
	return true;
}

if(!this.requireLoginForAdlDownload) {
	return userStatus;
}

return isLoggedIn;```

src/app/asset-page/asset-page.component.ts
crystalfinch: This was the value that Spencer from AUTH suggested, so I certainly hope so! 😅 It gave me what I expected when loading the site in various ways.

crystalfinch: Renamed the variable to "userAuthedWithInstitution" in ea5014d

src/app/asset-page/asset-page.component.ts
crystalfinch: Refactored in f30adbf

PR #1282, opened 2020-04-29
BLUE-260: Support web components in IE11

Niznikr:
Resolves BLUE-260

## Description

Angular (the cli specifically) expects your dependencies to be available in ES5 format if you intend on supporting IE11. The fun thing is they don't provide a way for you to transpile dependencies yourself in instances where the library is not available in ES5. Luckily there is a [tool](https://www.npmjs.com/package/@angular-builders/custom-webpack) that allows us to use custom webpack configs along with the cli. Also had to move where the web component polyfills were being applied as they were conflicting with the other polyfills AIW is using.

## Checks

- [ ] Yes
- [ ] Not yet
- [x] Not applicable

- [ ] Yes
- [ ] Not yet
- [x] Not applicable

**Browsers tested on:**

- [x] Chrome
  - [ ] Mobile
- [ ] Safari
  - [ ] Mobile
- [ ] Firefox
  - [ ] Mobile
- [ ] Edge
- [x] IE11
- [ ] Not applicable

### DIFF
--- a/browserslist
+++ b/browserslist
@@ -1,0 +1,5 @@
+> 0.5%
+last 2 versions
+Firefox ESR
+not dead
+IE 11
--- a/src/polyfills.ts
+++ b/src/polyfills.ts
@@ -2,10 +2,6 @@
 
 // import 'ie-shim'; // Internet Explorer 9 support
 
-// Web component polyfills for IE11
-// import '@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js';
-// import '@webcomponents/webcomponentsjs/webcomponents-loader.js';
-
 // import 'core-js/es6';
 // Added parts of es6 which are necessary for your project or your browser support requirements.
 import 'core-js/es6/symbol';
--- a/src/polyfills.ts
+++ b/src/polyfills.ts
@@ -3,8 +3,8 @@
 // import 'ie-shim'; // Internet Explorer 9 support
 
 // Web component polyfills for IE11
-import '@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js';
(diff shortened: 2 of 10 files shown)


browserslist
```diff
@@ -0,0 +1,5 @@
+> 0.5%
+last 2 versions
+Firefox ESR
+not dead
+IE 11
```
michael-iden, line 5: Hadn't seen this file before; seems like a cool tool. Can you describe what would happen when building, testing and deployment would look like if we didn't have this configuration

Niznikr: Sure thing! This just helps @babel/preset-env [know](https://babeljs.io/docs/en/babel-preset-env#browserslist-integration) how to transform code for supporting your desired browsers. In this case we need babel to transpile the web component related dependencies (Pharos, lit) to a format supported by IE, to prevent it from crashing in the browser.

NazimAli2017: @michael-iden `browserlist` is also used as part of the JSTOR assets build process, using the default configuration.

src/polyfills.ts
```diff
@@ -3,8 +3,8 @@
 // import 'ie-shim'; // Internet Explorer 9 support
 
 // Web component polyfills for IE11
-import '@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js';
-import '@webcomponents/webcomponentsjs/webcomponents-loader.js';
+// import '@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js';
```
NazimAli2017, line 6: Do we need the commented out lines?

Niznikr: Nah I can clean them!

PR #1290, opened 2020-06-22
PLATO-844: Port number was not showing up in sharing URL when connecting through a proxy

gbassham:
PLATO-844

## Description
Previously, we were only passing the hostname to the asset page to use for the sharing URL. This is a problem for proxies that need a port. We are now passing the port number as long as it is not 80, 4000, or 8080. We check the port number against an array to ensure that we do not pass those ports to the client.

I also refactored some existing code that would have displayed a bad URL if the port number were 80XX, but not 8080.

## Checks

- [ ] Yes
- [ ] Not yet
- [X] Not applicable

- [ ] Yes
- [ ] Not yet
- [X] Not applicable

**Browsers tested on:**

- [X] Chrome
  - [ ] Mobile
- [ ] Safari
  - [ ] Mobile
- [ ] Firefox
  - [ ] Mobile
- [ ] Edge
- [ ] IE11
- [ ] Not applicable


src/app/app.service.ts
```diff
@@ -18,20 +18,30 @@ export class AppConfig {
...
     this.isBrowser = isPlatformBrowser(platformId);
+
     // Identify hostname from request or client side
+    this.isBrowser = false;
     if (this.isBrowser) {
-      this.clientHostname = window.location.hostname
-    } else{
+      this.clientHostname = this.getHostName(window.location.hostname, window.location.port);
+    } else {
       let req = this.injector.get('request');
```
gbassham, line 31: I refactored this code to use the `getHostname` function because I thought there could be problems if we had a port number like `8001`.

src/app/app.service.ts
```diff
@@ -18,20 +18,30 @@ export class AppConfig {
   private _config
   private isBrowser
 
+  private PORTS_TO_EXCLUDE = ["80", "4000", "8080"];
+
   constructor(@Inject(PLATFORM_ID) private platformId, private injector: Injector) {
     this.isBrowser = isPlatformBrowser(platformId);
+
     // Identify hostname from request or client side
+    this.isBrowser = false;
```
eslawski, line 27:
```suggestion
```

gbassham: done

src/app/app.service.ts
michael-iden: I poked around a bit but I am generally confused why we would exclude any ports at all? Maybe something with the server side rendering?

src/app/app.service.ts
```diff
@@ -18,20 +18,30 @@ export class AppConfig {
...
+      if(req) {
+        const requestHost = req.get("host").split(":");
+        this.clientHostname = (requestHost.length === 1) ?
+                                requestHost[0] :
+                                this.getHostName(requestHost[0], requestHost[1]);
+      }
+      else {
+        this.clientHostname = "";
+      }
+
```
eslawski, line 40: I noticed that we have two conditions here to determine the host name, and it depends on if `isBrowser` is true or false? Not sure I am understand the use cases for both. Any idea what the difference is?

I guess I am mostly asking because it appears we only did the filtering of the ports if `isBrowser=false`. And honestly not too sure what that was designed for in the first place.

gbassham: It has to do with server-side rendering, and when `isPlatformBrowser` is false, the window variable is undefined, so I imagine that is the reason for the if/then statement.

src/app/app.service.ts
gbassham: Maybe? I didn't want to change the current functionality.

eslawski: It seems like we only filtered out the ports when dealing with the server side rendering? Wondering if we should narrow the scope to just that part of the if/else branch?

gbassham: No, the problem is when `isBrowser` is set to `true`. Before we were only passing `window.location.hostname` back, which was causing the share URL for proxies to not work. So at a minimum we need to pass back the port for the `isBrowser` section of the if/then. I filtered out the three ports so that you would get a consistent URL whether `isBrowser` was true or false.

The share URL actually should be fine most of the time when `isBrowser` is false because it only filters out those three ports. (So a port of 2443, which was the issue in the ticket, would be fine.) I changed that code just because if the port were, say, 8001, then it would return {hostname}01 instead of {hostname}:8001. It's probably an edge case but it seemed worth fixing.

gbassham: and I am assuming there was a specific reason for the original filtering out of 8080, 4000, and 80, but I do not know what that reason was.

eslawski: Gotcha. Ya I just didn't know if there was a reason we were not filtering out ports when `isBrowser` is true or something like that.

gbassham: we were actually filtering out all ports when `isBrowser` was true. But I don't know whether that was intentional or the original coder expected that `window.location.hostname` would return the port the way the `host` in `request` does.

jialin-he (approved):

eslawski (approved):

PR #1355, opened 2021-09-27
PALS-1149: role prompt in artstor 

adandavati:
Resolves PALS-1149

## Description

If a user is logged in (and doesn't have a role, and hasn't been prompted for one) we prompt them to enter their role. If they respond with a role we update their role and note that they were prompted. If they decline, we note that they were prompted so we don't ask them again.
Screen Shot 2021-10-01 at 1 58 51 PM


adandavati:
PACTs incoming!

src/app/modals/role/role.component.pug
```diff
@@ -0,0 +1,16 @@
...
+          span(aria-hidden="true") &times;
+      .modal-body.btn-group-vertical
+        p([innerHtml]="'You&apos;re a...' | translate")
+        a.btn.btn-secondary((click)="saveRole('ROLE_CU_UNDERGRAD_STUDENT')") {{ 'Undergraduate Student' | translate }}
+        a.btn.my-sm-1.btn-secondary((click)="saveRole('ROLE_CU_GRAD_STUDENT')") {{ 'Graduate Student' | translate }}
+        a.btn.my-sm-1.btn-secondary((click)="saveRole('ROLE_CU_FACULTY')") {{ 'Instructor/Faculty' | translate }}
+        a.btn.my-sm-1.btn-secondary((click)="saveRole('ROLE_PS_TEACHER')") {{ 'K-12 Teacher' | translate }}
+        a.btn.my-sm-1.btn-secondary((click)="saveRole('ROLE_CU_LIBRARIAN')") {{ 'Librarian' | translate }}
+        a.btn.my-sm-1.btn-secondary((click)="saveRole('ROLE_OTHER')") {{ 'Other' | translate }}
+        a.link.my-sm-1.btn-link((click)="dismissModal()") {{ 'No thanks' | translate }}
```
michael-iden, line 16: I don't know artstor very well but are we able to get the same UX with these underlying elements being buttons. They feel more like buttons rather than links that would perform some sort of navigation.

adandavati: @michael-iden only the last element, the no thanks link, is a link. Everything else is a button!

src/app/app.component.ts
```diff
@@ -268,6 +270,13 @@ export class AppComponent {
     this._fullscreen.addFullscreenListener();
   }
 
+  /**
+   * Determines whether to display role prompt
+   */
+  private initializeRolePrompt(shouldPromptForRole: boolean): void {
+    this.showRolePrompt = this._auth.shouldPromptForRole() && shouldPromptForRole
```
michael-iden, line 277: Ultra nit but these feel a bit backwards to me, assuming this would short circuit to false if the `shouldPromptForRole` flag value is false.

adandavati: @michael-iden like this? 
`this.showRolePrompt = shouldPromptForRole && this._auth.shouldPromptForRole()`

src/app/modals/role/role.component.pug
michael-iden: I'm not super familiar with pug but I'm reading this as these elements are all links (anchor HTML tags) but styled to look like buttons in some cases. From an accessibility standpoint, I think we've gotten dinged on HTML elements that are styled to look like other HTML elements since sometimes they read out differently with screen readers and may have slightly different keyboard controls.

michael-iden: image

adandavati: I think you're right, thanks for clarifying. I'm not super familiar with pug either.

src/app/app.component.ts
michael-iden: You got it. sry I should've added the suggestion from the start
```suggestion
    this.showRolePrompt = shouldPromptForRole && this._auth.shouldPromptForRole()
```

adandavati: No worries! Changed.

michael-iden (approved):

ymouzakis (approved):
Nice work!

src/app/modals/role/role.component.pug
```diff
@@ -0,0 +1,16 @@
+.modal.fade-in.show
+  .modal-dialog(role="document")
+    .modal-content
+      .modal-header
+        h4.modal-title([innerHtml]="'Complete your Profile' | translate")
```
daneah, line 5: Does this follow our typical capitalization rules? I think in JSTOR we'd use "Complete your profile." Here it seems like it should be at least "Complete Your Profile," if not "Complete your profile."

adandavati: Updated!

PR #1392, opened 2022-06-30
PALS-1482: Build Post Login AJI banner on AIW

ssharma2ithaka:
PALS-1482

## Description

Post login banner implementation.

image


ssharma2ithaka:
image

adandavati (changes requested):
@ssharma2ithaka I believe we want the current skyBanner to continue to function as it used to. I am concerned these changes will break/prevent that.

ssharma2ithaka:
> 

@adandavati I can leave the previous skyBanner up and create a new component for post login banner if that works?

adandavati:
@ssharma2ithaka is this banner visible for all users? Does it only pop up after the AJI intercept has been dismissed?

ssharma2ithaka:
> @ssharma2ithaka is this banner visible for all users? Does it only pop up after the AJI intercept has been dismissed?

Yes, as of now the banner is visible to all users. It actually pops up as soon as the page is loaded.

adandavati:
@ssharma2ithaka before we promote this to test we will need to prevent the banner from being visible to all, and gate it the same way as the modal.

PR #1402, opened 2022-08-02
PALS-1534: Update the Artstor AJI Modal/Banner with copy, and a list of users to roll out to

ssharma2ithaka:
Resolves PALS-1534

- Update logic for displaying AJI intercept and post login banner to targeted users
- Add data-sc attributes for GA tracking

AJI Intercept
image

Post-login drop down banner
image

Link behavior on hover
image

PR #1403, opened 2022-08-05
PALS-1534: continues from https://github.com/ithaka/aiw-ui/pull/1402

ssharma2ithaka:
Resolves PALS-1534

- Continues https://github.com/ithaka/aiw-ui/pull/1402

### DIFF
--- a/src/app/_services/auth.service.ts
+++ b/src/app/_services/auth.service.ts
@@ -437,7 +437,7 @@ export class AuthService implements CanActivate {
     let ajiInterceptWasClosed = this.getFromStorage('AJIInterceptClosed')
     // let userIsLoggedInJstorUser = user && user.isLoggedIn && institution.institutionName && institution.institutionName == 'JSTOR'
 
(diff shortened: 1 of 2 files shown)


src/app/_services/auth.service.ts
```diff
@@ -424,7 +424,7 @@ export class AuthService implements CanActivate {
     if (this.getFromStorage('AJIInterceptClosed')) {
       return false
     }
-    return user && user.isLoggedIn && user.preferences.hasOwnProperty('showAJIModalOrBanner') && user.preferences.showAJIModalOrBanner
+    return user && user.isLoggedIn && user.hasOwnProperty('showAJIModalOrBanner') && user.showAJIModalOrBanner == "true"
```
adandavati, line 427:
```suggestion
    return user && user.isLoggedIn && user.hasOwnProperty('showAJIModalOrBanner') && user.showAJIModalOrBanner
```

adandavati (approved):

src/app/_services/auth.service.ts
adandavati: ^ Ignore this suggestion. I think checking against "true" is better in this case.

PR #1439, opened 2023-02-16
PALS-1760: update the password reset workflow

adandavati:
Change the mechanism for updating password. 

Created a new page at /#/password/reset


src/app/password-reset/password-reset.component.ts
```diff
@@ -0,0 +1,99 @@
...
+    this._auth.validateSignature(this.passwordResetParams.email, this.passwordResetParams.signature).then( res => {
+      if (res !== true) {
+        this.invalidSignature = true
+      }
+    })
+  }
+
+  submit() {
+
+  }
```
kentjas1, line 61: Is this empty method needed?

kentjas1 (approved):
One question about an empty method but with my limited knowledge of Angular, nothing standing out.

PR #1443, opened 2023-03-24
PALS-1796: migrate aiw to capstan/gitlab

adandavati:
Create a gitlab pipeline to deploy aiw-ui that removes the dependency on local machines. 

- switch from npm to yarn
- update status page to resolve ci build error
- update appdynamics to a package that is more compatible
- remove the dependencies on a users local machine in the deploy process

Screen Shot 2023-03-31 at 1 43 02 PM


shoupeva-ithaka (approved):

afiasco (approved):

mtorres3 (approved):

Dockerfile
```diff
@@ -10,8 +10,9 @@ COPY dist ./dist
 # Copy server package.json for commands and dependencies
```
michael-iden, line 10: Using node 11 up above in this Dockerfile is somewhat concerning. Probably out of scope for this migration, and may be lower hanging fruit for whenever a team can look at this next. Thanks for jumping on this app and getting it moved to K8S!

NazimHAli: Yup, we kept the same versions and only upgraded a few packages because of installation issues in the CI. We can look into upgrading node in the future and most of the packages are out of date (some >5 years).

PR #1458, opened 2024-06-21
Add flag for redirection

daneah:


src/app/_services/jstor.redirect.service.ts
```diff
@@ -1,24 +1,56 @@
...
+    const doRedirect = enabledFlags.includes(REDIRECT_FLAG);
+
     const currentRequest = window.location.href;
 
     if (currentRequest.includes('/#/')) {
       const params = new URLSearchParams({artstorPath: currentRequest }).toString();
       fetch(`/get-the-redirect-please/?${params}`).then(async resp => {
-        // TODO: Logic to handle "location" response from server and 301 user to that
         const data = await resp.json();
         if (data.location) {
```
michael-iden: This is a little out of scope but do we know what we want to do if `data.location` evaluates to false? Because in that case, the user will continue to hang out on artstor? Do we eventually need to add a fallback?

kentjas1 (approved):
:shipit:

src/app/_services/jstor.redirect.service.ts
daneah: I think it is the server-side responsibility to do this, and don't want to get overly defensive here—TBD, I think.

michael-iden (approved):

NazimHAli (approved):

src/main.ts
```diff
@@ -33,9 +33,64 @@ import { environment } from './environments/environment';
...
+})
+  .then(response => response.json())
+  .then(flagData => {
+    const flagStates = flagData.data ? flagData.data : null;
+    const enabledFlags = flagStates ? flagStates.flags.enabled : [];
+    const doRedirect = enabledFlags.includes(REDIRECT_FLAG);
+
+    const currentRequest = window.location.href;
+
+    if (currentRequest.includes('/#/')) {
```
eslawski: I think we might have broken any urls that don't have the `/#/` because we never end up calling `initializeApp()` if this condition fails? I think this is evidenced by the confusion Mike and I ran up against when trying to load `https://stage.artstor.org` vs `https://stage.artstor.org/#/`. 

The only other url I can think of without the hash are the `/public` ones, and I am curious how those ones are going to work now for the next month and a half. I know they are server side rendered, but they might get tripped up now too by this condition when they go to render server side

Could a solution be to just add an `else` condition to this that also calls `initializeApp()`?

daneah: I think you're right, yep.

kentjas1 (approved):
:shipit:

michael-iden (approved):

eslawski (approved):

